PR #6, opened 2022-03-23
chore: add multi-container sessions

olevski:


rokroskar:
I really like this idea - I think we will need this sort of functionality sooner rather than later (see #11). We should complete and merge this RFC and work on defining the apps/dashboards use-case. Serving multiple containers could imply, for example, multiple Dockerfiles in a project, it can complicate CI, etc.

design/005-multi-container-sessions/005-multi-container-sessions.md
```diff
@@ -0,0 +1,163 @@
...
+      livenessProbe: {}
+      readinessProbe: {}
+      startupProbe: {}
+  routing:
+    host: dev.renku.ch
+    ingressAnnotations:
+      kubernetes.io/ingress.class: nginx
+      nginx.ingress.kubernetes.io/proxy-body-size: "0"
+      nginx.ingress.kubernetes.io/proxy-buffer-size: 8k
+      nginx.ingress.kubernetes.io/proxy-request-buffering: "off"
```
rokroskar, line 109: maybe the RFC doesn't need this level of detail?

design/005-multi-container-sessions/005-multi-container-sessions.md
```diff
@@ -0,0 +1,163 @@
...
+    path: /sessions/<session_name>
+    tls:
+      enabled: true
+      secretName: dev-renku-ch-tls
+  storage:
+    pvc:
+      enabled:
+      mountPath: /home/jovyanl/work
```
rokroskar, line 117:
```suggestion
      mountPath: /home/jovyan/work
```

the `mountPath` might need to be set per container?

design/005-multi-container-sessions/005-multi-container-sessions.md
```diff
@@ -0,0 +1,163 @@
...
```
rokroskar, line 59:
```suggestion
      mountPath: /home/jovyan/work
```

olevski:
@rokroskar I am sorry for the delay. I will clean this up and request another review so that we can merge it.

design/005-multi-container-sessions/005-multi-container-sessions.md
olevski: I removed the ingress annotations. I agree. Not sure if that is what you were referring to.

rokroskar: yes, that's it :)

design/005-multi-container-sessions/005-multi-container-sessions.md
olevski: I was thinking that we just mount the pvc in the same place on all containers. This is the spec for the jupyterserver so it does not have to be a valid pod spec.

I just thought it was easier to enforce this rather than to have freedom to mount things in different locations.

olevski:
@rokroskar all done let me know what you think

design/005-multi-container-sessions/005-multi-container-sessions.md
olevski: But it is definitely doable to have more freeedom and mount thing differently across different "apps"/containers.

rokroskar: The problem is that this assumes a jovyan user exists - which potentially complicates the requirements of the image. But maybe it's a non-issue? not sure.

olevski: Ok yeah I will add this to the questions that should be resolved then.

It will be nice to not have to worry about this.

rokroskar: still, "jovyan" is misspelled ;)

rokroskar (approved):

PR #8, opened 2022-03-30
feat: add RFC about external data

rokroskar:
Description of how we could potentially handle external data sources in Renku.


ciyer:
This RFC contains many details about implementation, but I still am finding it difficult to imagine how it would be used in practice.

As someone who uses AWS S3, it would be really great to have something like:

```
renku dataset add --name [name of the dataset] --external s3://[bucket]/[path]
```

What I would expect this to do is to:
1. create a dataset
2. if the path is to a file, the dataset contains just the file; if the path ends in a `/`, the dataset contains, recursively, everything below and including the path 
3. If I have the aws cli set up, the command should do an `aws s3 cp [--recursive] [s3-url]` to get the contents into my project

A nice extension would be the following: 

```
renku dataset add --name [name of the dataset] --external s3://[bucket]/[path] --local-path [some location where the contents should be stored]
```

When I work with s3, the contents there are often very large, and I do not want my local file system. In those cases, I would put them on an external disk or NAS, or, I guess, FUSE-mount the contents as a new volume. But, I would like to have all of those options and as a user be able to decide if I want the S3 content brought in to: a folder of the project, an external volume, or FUSE-mounted volume

rokroskar:
Thanks @ciyer - yes, this last one was exactly what I had in mind as the default. But maybe there are use-cases where copying data into the project and using git-LFS from there (as is done now for e.g. grabbing data from zenodo) would still be desired? 

imo the `--external` flag is not needed because we can guess that from the URL.

rokroskar:
One more clarification - the reason I thought we might need a new command such as `stage` is because the local path to that fuse-mounted S3 bucket might be different from system to system. I can imagine we'll need a way to manage/query that, which requires new functionality and interface.

ciyer:
When I use s3, I do it in one of two ways:

**Amount of data is small**
- write a script to fetch the data and store it in a subfolder of the project
- add the script to git
- add the subfolder to gitignore

This way, if I clone the project to a new computer, I can easily retrieve the data again. I do not add the files to LFS (or, in my annex) because I can always retrieve them again if necessary, and I trust them to be available from S3 at a future data.

**Amount of data is large**
- write a script to fetch the data and store it on an external drive
- add the script to git
- write scripts that reference the external drive and reduce the data to a manageable size
- add those scripts to git
- if the reduced data is very small, I will add it to git as well, if it is too large, I will put it in a gitignored subfolder

In this case, if I clone the project to a new computer, I either have the reduced data that I need for day-to-day work available, or, I plug in the external drive and re-run the scripts to re-compute the reduced data.

rokroskar:
@ciyer so I think what I proposed would basically replace the need to write one-off scripts for managing the data in S3, in the way that you have described.

design/004-external-data-handling/004-external-data-handling.md
```diff
@@ -0,0 +1,232 @@
...
+External data has to somehow be brought into the system where Renku is running.
+This seems to be a "staging" operation, but we could also call it something
+else. "Mounting" is another option but seems too linux-specific.
+
+```
+renku dataset stage very-large-data
+```
+
+This command uses Fuse when necessary to mount data locally. Once it's staged,
+it can be found locally and used in workflows:
```
Panaetius, line 181: Well since cwl (and toil?) copy files to a temporary directory for execution, we would also need to handle that properly, linking instead of copying or submitting absolute paths.

design/004-external-data-handling/004-external-data-handling.md
```diff
@@ -0,0 +1,232 @@
...
+
+```
+renku dataset add s3://server/path/to/bucket
+Access Key: 1234
+Secret Key: abcd
+Store credentials? (y/N): y
+Credentials stored in ~/.renku/renku.ini
+```
+
+`s3://` is used here but we should also support at least `nfs://`. Renku should
```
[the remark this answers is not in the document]
rokroskar: yes, definitely. I meant to imply that by the `Provider interface` section, but it should probably be made more explicit :)

design/004-external-data-handling/004-external-data-handling.md
rokroskar: are there built-in mechanisms in toil and/or cwl to do this automatically? @vigsterkr maybe you know?

design/004-external-data-handling/004-external-data-handling.md
```diff
@@ -0,0 +1,232 @@
...
+#### Streaming external data
+
+The renku-python API should offer some support for more easily accessing remote
+data, especially for sources from which it is common for data to be processed as
+a stream. Ideally, Renku could lazily stage the external data if access to the
+resource is requested.
+
+## Drawbacks
+
+* Relying on fuse is tricky because it requires elevated privileges on all
```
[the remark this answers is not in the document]
rokroskar: @olevski do you have any insights on this?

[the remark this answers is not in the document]
Panaetius: so locally you don't need escalated privileges, as long as you have access to `/dev/fuse` and in an interactive session, we could do it in an escalated sidecare with a microservice.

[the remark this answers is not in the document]
olevski: I do not have really much experience with using fuse on personal laptop. I thought that you do need sudo but I guess that is not always the case. I just did `sshfs` on my mac and that did not require sudo.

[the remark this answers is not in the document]
olevski: IMO this feels a really big issue to tackle. Can we split it up in two pieces:
1. How to track provenance of external data
2. How to mount nfs/s3/etc with the cli

I think it may be a bit easier to think of these separately.

Also trying to get all this mounting to work on 2/3 operating systems seems very intimidating.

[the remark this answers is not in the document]
olevski: I got some background and more info from @Panaetius on this. And I feel better now :)

olevski:
One more thing I wanted to add is that using the cli in a [csi driver](https://kubernetes-csi.github.io/docs/) written specifically for Renku would allow us to use the same code for mounting stuff locally and in k8s. The go code from the csi driver would call the renku cli to mount stuff.

I have been trying to write a simple csi driver for mounting s3 buckets with s3 fuse (mostly on the side) I just have not had time to finish this.

With a csi driver we get a storage class (for example renku-storage) and then we can create a PVC (that uses said storage class) that will result in a s3 bucket, database or whatever the cli supports being mounted in a session. So if we write our own csi driver we would not have to use a privileged sidecar to run the renku cli and mount stuff in a session in k8s.

rokroskar:
> One more thing I wanted to add is that using the cli in a [csi driver](https://kubernetes-csi.github.io/docs/) written specifically for Renku would allow us to use the same code for mounting stuff locally and in k8s. The go code from the csi driver would call the renku cli to mount stuff.

I wanted to ask you about this! I imagined it could work the way that you describe but I wasn't sure. The potential benefit of a sidecar is that we could (I think) make it mount volumes dynamically... whereas with the PVC we would always need to recreate the pod.

olevski:
> The potential benefit of a sidecar is that we could (I think) make it mount volumes dynamically

That is true. It is worth asking then the BIT how much of a problem such a feature will be. They want to limit the number of stuff that runs as root. And I think that this would require root / elevated permissions.

I can send them a message just to get a sense.

rokroskar:
Root is not necessary, but the container will need to be run with elevated privileges afaik.

olevski:
I had some manifests to test mounting stuff from a container sidecar. I can dig them up and confirm exactly what security context is needed to run this. It would be good to know and have this yaml snippet.

design/004-external-data-handling/004-external-data-handling.md
```diff
@@ -0,0 +1,255 @@
...
+need to envision something new like `renku data source` which the `dataset`
+commands could reference.
+
+
+```
+renku dataset add s3://server/path/to/bucket
+Access Key: 1234
+Secret Key: abcd
+Store credentials? (y/N): y
+Credentials stored in ~/.renku/renku.ini
```
lokijuhy, line 171: If the credentials are saved in ~/.renku/renku.ini, does this mean that anyone who forks the repo has the credentials?

rokroskar: no that's in the user's home directory

lokijuhy:
Does this RFC intend to support reading and writing to external data systems, or only reading?

design/004-external-data-handling/004-external-data-handling.md
lokijuhy: Oh right. Does that mean the credentials are wiped with each RenkuLab session, i.e. if I enter the credentials in one RenkuLab session, then start a new session a few days later, I have to enter my credentials again?

rokroskar:
So far it was talking about _sources_ specifically. Not sure if we should include writing as well...

design/004-external-data-handling/004-external-data-handling.md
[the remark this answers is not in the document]
rokroskar: @Panaetius I'd be curious to know what this comforting information was... :)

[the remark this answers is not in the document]
lokijuhy: > I got some background and more info from @Panaetius on this. And I feel better now :)

@olevski I have the same feeling- what did you learn that made you feel better?

design/004-external-data-handling/004-external-data-handling.md
```diff
@@ -0,0 +1,255 @@
...
+
+* Relying on fuse is tricky because it requires elevated privileges on all
+  systems. On linux, the user must have sudo and it isn't clear how this will
+  work at all on windows. Fuse has the potentially to work very seamlessly but
+  can also cause problems. *EDIT*: this turns out not to be an issue; outside of
+  docker containers FUSE does not require additional privileges.
+
+* It might also be inefficient to keep track of which data is actually external
+  when recording/handling workflows.
+
```
lokijuhy, line 247:
```suggestion

## Out of Scope
* Saving credentials, so that RenkuLab users don't have to re-enter their credentials at the start of each session.

```

design/004-external-data-handling/004-external-data-handling.md
```diff
@@ -0,0 +1,255 @@
...
+Sally is Bob's collaborator who continues to build on Bob's initial work by
+building a more robust workflow. She uses standard Renku commands to record the
+workflows. Renku uses the same mechanisms to attach the data as described above
+when workflows are being re-executed or managed. She can also request to check
+whether the results are still in sync with the data using `renku status`. In
+this case, Renku checks the object hashes that were recorded previously against
+the data currently found in the remote location (this might take a while over
+the network - some data sources might provide pre-computed hashes). If the
+hashes differ, Renku offers to update the results.
+
```
lokijuhy, line 71:
```suggestion

#### 3. Data storage for research project

Willow is interested in a large, protected dataset that is offered for free to researchers by an external organization. She has to accept a data access agreement to access the data, at which points she receives credentials to the bucket. To explore the dataset, she creates a Renku project and mounts the S3 bucket using the credentials.

After some exploration of the dataset, Willow decides to start a research project with the dataset. She makes a copy of the dataset in her own private S3 bucket for two reasons: (1) to make sure she maintains access to the data independent of the external organization, and (2) because she prefers to keep the project’s raw, intermediate, and final data products in the same place. She creates a Renku project for her work and mounts the bucket with her credentials. Throughout the research project, she reads the raw data from the mounted directory and saves intermediate data files and figures in separate folders in the same S3 bucket.

```

lokijuhy:
> So far it was talking about _sources_ specifically. Not sure if we should include writing as well...

I added a third user story for a researcher project use case, which motivates the need to write to the data system as well. I think the use cases that only need read-access are fairly limited.

rokroskar:
> I think the use cases that only need read-access are fairly limited.

That probably depends on the domain. There are plenty of cases where large raw data needs to be made accessible but downstream data products will be stored separately and be much smaller in size.

design/004-external-data-handling/004-external-data-handling.md
rokroskar: indeed there is a `--linkImports` option that will do a hard-link. But my guess is that this won't really work with fuse-mounted fs.

ciyer:
> 1. use data from remote storage as-is by remote-mounting it and remembering the source in renku metadata. This should support reading and writing.
> 2. use external storage as a source/sink but make local copies to speed up access for iterative calculations. In principle, this should also support reading and writing.

I think this would work. 👍 Having local copies (2) is necessary to support use outside renkulab, though I think that just reading in that case is an acceptable limitation.

rokroskar:
Why do you say it's necessary to make local copies to use it outside of renkulab? We should be able to do mounts also on at least macos and linux systems.

lokijuhy:
@rokroskar I think clarifying the 2 use cases is helpful, thank you! To summarize Use Case 1:
* User can read and write to a mounted external file source (with expected network performance penalties)
* User will be able to view metadata for data that came from an external file source: url, perhaps s3 bucket name... 
* User will be able to run `renku run/update/rerun` as with local data sources, where the data in the external file source is pulled/checked for updates, likely with network performance penalties (is this right?)

And I second your question to @ciyer about why (2) is necessary for outside renkulab/`renku session`?

ciyer:
Good point -- if you make the assumption that FUSE is available on the user's machine, then you could enable access to S3 with just use-case 1, but that would not cover my use case detailed above.

I need to be certain that I can access the data even if I am not connected to the internet or if the bucket owner fails to pay their AWS bill, rendering the S3 bucket inaccessible ;) . I have not looked into the the details of how the FUSE/S3 file system is implemented, but I suspect that the data would not be accessible in those cases, and in any case, I like the security of knowing that I have a straightforward local copy of the data.

rokroskar:
@ciyer yes, I think what you describe is covered by my use-case 2

design/007-external-data-handling/004-external-data-handling.md
```diff
@@ -0,0 +1,421 @@
...
+
+#### 4. Access remote data but make local copies
+
+Sharing data in remote S3 buckets might not be only for large data that cannot
+be stored locally; in some cases it might be prudent to make local copies. Users
+frequently write their own supporting code for handling this.
+
+##### In the case where the data is small:
+
+1. write a script to fetch the data and store it in a subfolder of the project
```
olevski, line 98: Writing this script is what we expect the users to do? Or what we intend to do?

design/007-external-data-handling/004-external-data-handling.md
```diff
@@ -0,0 +1,421 @@
...
+### Use-cases
+
+From the user stories, two primary high-level use-cases emerge:
+
+1. use data from remote storage as-is by remote-mounting it and remembering the
+   source in renku metadata. This should support reading and writing.
+
+2. use external storage as a source/sink but make local copies to speed up
+   access for iterative calculations. In principle, this should also support
+   reading and writing.
```
olevski, line 126: So we need some kind of a system/method for avoiding conflicts when local data is written to, right? If two people mount a large dataset and download the data locally around the same time what is the final version that is in the S3 bucket at the end? When are changes reflected in the S3 bucket when a user makes changes locally? For example, do we sync immediately or do we depend on the user for syncing or do we sync at specific times (i.e. when the user stops their session)?

Is this something we should discuss here in the RFC or not?

design/007-external-data-handling/004-external-data-handling.md
```diff
@@ -0,0 +1,421 @@
...
+Provider is to configure the project and/or the user's environment in such a way
+as to make the `DataSource` available in a consistent, repeatable location with
+respect to the project and without boilerplate.
+
+### Local vs. hosted interactive sessions
+
+Relying on FUSE for providing filesystem access to external resources means that
+locally we do not require special privileges and can manage the data access for
+the user. In the docker-based remote sessions, the situation is more complicated
+because containers require elevated privileges to access FUSE on the host.
```
olevski, line 220: This could be a problem for hardened/enterprise deployments that have strict security policies that are not allowed to be changed or require a lot of effort so that we can get permission to deviate from them.

design/007-external-data-handling/004-external-data-handling.md
```diff
@@ -0,0 +1,421 @@
...
+-- output --
+Name: very-large-data
+Created: 2022-03-30 15:55:59+02:00
+Creator(s): Rok Roškar <rok.roskar@sdsc.ethz.ch>
+Title: very-large-data
+Description: Data that is very large. Indeed.
+Data Sources:
+    - type: s3
+      url: s3://server/path/to/bucket
+      location: None
```
olevski, line 271: Does this mean the data has not been staged anywhere? And `renku dataset stage ...` is required so that the data is actually available somewhere on the file system.

design/007-external-data-handling/004-external-data-handling.md
```diff
@@ -0,0 +1,421 @@
...
+
+```
+renku storage stage --all --copy
+```
+
+Note that in the case of remote storage (as opposed to datasets) there is no
+contradiction in writing directly to the directory where the data source is made
+available in the local project. If the data source is mounted, then the changes
+are reflected on the remote location immediately; if the data is copied, it
+should be possible to implement a `sync` command that would copy upstream new or
```
olevski, line 379: What happens when we have conflicts during this sync operation? I.e. when a remote file is newer than a local file? Also if one person has the same dataset mounted with fuse (i.e. without copying) do we allow another person to add the data with either the `--copy` flag or without.

olevski: A few things that are somewhat related are:
1. Handling buckets to which the user has no write access. In this case we should make sure that after mounting the files (with or without `--copy`) the user is also not allowed to write to them
2. We should have a flag in the `storage`/`dataset` `add` command to indicate whether a dataset is read-only or read/write. And this should be compatible with the access level that the credentials provide. 
3. There may be tricky corner cases with granular access control policies that I think in the case of S3 can apply only to specific files or subfolders.

design/007-external-data-handling/004-external-data-handling.md
rokroskar: Hmm that's a good question. I'm not sure this needs to be our responsibility tbh. If I think of how it works on other shared infrastructures, the infrastructure provider isn't making sure that users don't overwrite each other's stuff in shared directories.

design/007-external-data-handling/004-external-data-handling.md
rokroskar: yes - dynamic mounting might have to be a feature that you can turn off if needed.

design/007-external-data-handling/004-external-data-handling.md
rokroskar: that was my initial thinking. But maybe staging should be the default and you could use a flag to avoid.

design/004-external-data-handling/004-external-data-handling.md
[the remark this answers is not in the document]
Panaetius: I'd also like to know!

design/007-external-data-handling/004-external-data-handling.md
```diff
@@ -0,0 +1,421 @@
...
+
+The renku-python API should offer some support for more easily accessing remote
+data, especially for sources from which it is common for data to be processed as
+a stream. Ideally, Renku could lazily stage the external data if access to the
+resource is requested.
+
+## Drawbacks
+
+* Relying on FUSE is tricky because it requires elevated privileges on all
+  systems. On Linux, access to /etc/fuse and `CAP_SYS_ADMIN` is needed for fuse to work.
```
Panaetius, line 392: I believe you mean /dev/fuse ? Also CAP_SYS_ADMIN is only for docker. Or do you mean for setting up fuse?

design/007-external-data-handling/004-external-data-handling.md
```diff
@@ -0,0 +1,421 @@
...
+-- output --
+<bucket> added as external storage
+```
+
+This would not immediately do anything with the data, but only create the
+`DataStorage` metadata object. For bringing in the data, the user would need
+to do
+
+```
+renku storage stage --copy <bucket> path/to/bucket
```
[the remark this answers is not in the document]
rokroskar: The data is copied to `.renku/mounts/<bucket>` - though arguably it should be called something other than `mounts`. Without `--copy` it's mounted there directly. In both cases it's symlinked to the location in the project. But maybe that's overcomplicating it and it could just be copied/mounted directly in `path/to/bucket`?

design/004-external-data-handling/004-external-data-handling.md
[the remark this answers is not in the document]
olevski: Lol it was mostly that we would try to implement the cli and notebook service components separately

design/007-external-data-handling/004-external-data-handling.md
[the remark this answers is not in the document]
Panaetius: I mean, the whole `Using external data _without_ Renku datasets` section uses `--copy` in all the examples, doesn't mention copying in the description around those commands, and does not address that a user could also access the mount directly without a copy.

lokijuhy:
@rokroskar Is this an accurate description of the different modes of data access we have to consider, and the current suggestions/contenders?

I have ... via a ...  | small data | big data
-- | -- | --
**renkulab session** | mount (FUSE) | mount (FUSE)
**local session** | copy OR install FUSE | access remote programmatically OR install FUSE |   |

rokroskar:
Apologies @lokijuhy @ciyer @Panaetius and @olevski but I reworked significant sections of this based on feedback and discussions we had over the past few days. Most importantly, it seems to me that copy as default makes the most sense and will lead to least friction. otoh we might also want to support some way of simple designating some external data as being used but never actually copying or mounting it (this is what "Mode 3: reference" refers to). Thanks for all the comments and feedback so far, I think it's slowly becoming clearer.

design/007-external-data-handling/004-external-data-handling.md
rokroskar: i've rephrased this bit slightly

design/007-external-data-handling/004-external-data-handling.md
```diff
@@ -0,0 +1,421 @@
...
+
+```
+renku dataset stage --copy very-large-data
+```
+
+The default location is to be `.renku/mounts/<name>` but can be specified with
+the `--location` flag in case it is preferred to take the data to an external
+drive or another filesystem.
+
+Once it's staged, it can be found locally and used in workflows:
```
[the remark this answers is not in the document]
rokroskar: tbh I didn't think about staging a dataset more than once - what would be the use-case?

design/007-external-data-handling/004-external-data-handling.md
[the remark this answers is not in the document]
rokroskar: I've reworked the copy/no-copy parts

design/007-external-data-handling/007-external-data-handling.md
```diff
@@ -0,0 +1,561 @@
...
+renku dataset stage external-dataset
+```
+
+This command copies by default. If the user wants to avoid copying and stage by
+mounting, FUSE can be used. For some filesystems we might be able to warn the
+user that the data they are trying to stage is very big and they should use the
+`--no-copy` flag instead.
+
+```
+renku dataset stage --no-copy external-dataset
```
lokijuhy, line 409: I think we should we be more explicit here, and make this flag `mount` or even `fuse-mount`, so it's clear what's happening (rather than what's _not_ happening)

design/007-external-data-handling/007-external-data-handling.md
```diff
@@ -0,0 +1,561 @@
...
+      location: None
+```
+
+Note that `location: None` because the dataset has not been staged yet. See
+below for details on that.
+
+#### Staging external data:
+
+External data has to somehow be brought into the system where Renku is running.
+This seems to be a "staging" operation, but we could also call it something
```
lokijuhy, line 396: What about "load"?

design/007-external-data-handling/007-external-data-handling.md
rokroskar: "fuse" is probably too jargon-y but maybe "mount" works?

design/007-external-data-handling/007-external-data-handling.md
rokroskar: then it would be, e.g. `renku storage load --mount`? Seems like "load" and "mount" mean two different things.

lokijuhy: I'm  not sure what you mean- is it bad that "load" and "mount" mean two different things? Loading via copying, vs loading via mounting.

rokroskar: yes to me it seems that you would either load or mount the data, in which case `load --mount` doesn't make sense. But maybe `load` is less confusing than `stage`.

lokijuhy: Mmm yes, I see what you mean. `stage` makes me think of git, and there, `stage` means preparing to send _out_ (`commit` and `push`), whereas in this context we mean more of a pull _in_. ... perhaps we could use `pull`?

design/007-external-data-handling/007-external-data-handling.md
```diff
@@ -195,6 +195,46 @@ openBIS etc. From the user's perspective, they all will generally work in the
...
+
+#### Kedro - uses programmatic access
+
+Kedro's data access is based on configuration files ([Data Catalog](https://kedro.readthedocs.io/en/stable/data/data_catalog.html#the-data-catalog)).
+In the configuration, kedro uses [fsspec](https://kedro.readthedocs.io/en/stable/data/data_catalog.html#specifying-the-location-of-the-dataset)
+to provide a standardized interface to remote data systems.
+Since kedro is python only, data is accessed programatically.
+In other words, the units of kedro workflows are python functions that take as
+input live python objects (dataframes, etc.), not file paths. fsspec helps map
+a variety of remote data locations to live python objects.
```
rokroskar, line 236: but then it doesn't copy and only does remote access?

lokijuhy: Correct. If you are referring to my statement in the PR comments that copying files is a "standard" approach, I meant that 2/3 systems I looked at use this strategy, though Kedro does not. Looking deeper into fsspec, its default is remote access, though it does support [local caching](https://filesystem-spec.readthedocs.io/en/latest/features.html#caching-files-locally) if you want.

design/007-external-data-handling/007-external-data-handling.md
```diff
@@ -0,0 +1,601 @@
...
+
+```
+renku dataset add s3-backed-dataset /path/to/file/on/filesystem
+```
+
+This adds a file from the local filesystem to the dataset. It is _copied_ to
+`data/s3-backed-dataset`. (Should it also be pushed to the remote storage
+immediately?)
+
+#### Adding external data to a Renku Dataset
```
mohammad-sdsc, line 390: Should we allow adding data from an external storage to an already-existing dataset? In other words, can a dataset contain data from multiple external sources? I don't see any reason not to do that. However, in this case we need to specify a data source (to push data) when we add local data to a dataset (e.g. an output of a computation).

rokroskar: This is a good point. 

> However, in this case we need to specify a data source (to push data)

You mean if a dataset has multiple sources defined, then on push you would have to say which destination should receive the newly added files?

design/007-external-data-handling/007-external-data-handling.md
```diff
@@ -0,0 +1,601 @@
...
+respect to the project and without boilerplate.
+
+Three scenarios are described here where a provider should ease the access to
+external data _and_ work with Renku internals to record information about
+provenance in the knowledge graph.
+
+#### Mode 1 (default): copy
+
+The default behavior of the CLI when staging the data locally (i.e. not on a
+Renkulab instance) should be to copy the data into the project. When disk space
```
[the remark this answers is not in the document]
rokroskar: That's a good question - it probably will need to be some action triggered by the user.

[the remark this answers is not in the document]
rokroskar: maybe a `renku push` command?

design/007-external-data-handling/007-external-data-handling.md
```diff
@@ -0,0 +1,601 @@
...
+principle, be added to a renku dataset with the "external" mechanism described
+above, but no metadata about its actual origins (i.e. an S3 bucket) would be
+recorded.
+
+#### Upcoming
+
+We have plans to include other external data providers, such as NFS shares,
+openBIS etc. From the user's perspective, they all will generally work in the
+same way as S3; data will be mounted to a local directory outside the repository
+and made accessible as network-mounted storage.
```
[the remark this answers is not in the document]
rokroskar: oh I just realized why the confusion... this section is merely talking about what has been discussed as possible features. The notion here about mounting is the same as the current S3 mounts in interactive sessions - it doesn't really have anything to do with the project, it's just made available somewhere in the session (outside of the repository).

design/007-external-data-handling/007-external-data-handling.md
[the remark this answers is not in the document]
rokroskar: I've added a section about this under "Commands"

rokroskar:
@lokijuhy @ciyer @Panaetius @mohammad-sdsc @olevski I believe I've addressed all the comments and concerns that have come up during the various discussions. It would be great if we can finalize this (for now) and start to think about next steps.

olevski (approved):

ciyer (approved):
This feature sounds great, and this explains it well.

mohammad-sdsc (approved):
Thank you!

design/007-external-data-handling/007-external-data-handling.md
rokroskar: @lokijuhy I suggest we leave this as is for now since we can't seem to find (agree on) a better alternative. I suspect that the true meaning/naming will crystalize once we actually try to use such a functionality.

lokijuhy (approved):
🎉

Panaetius (approved):

PR #9, opened 2022-04-28
feat: custodian s3 RFC

olevski:
This is an outline of how we can use the Custodian to inject s3 credentials in requests for launching notebook sessions for the specific project we discussed.

Please feel free to comment and to commit to this branch.

### DIFF
--- a/design/005-custodian-s3-credentials/005-custodian-s3-credentials.md
+++ b/design/005-custodian-s3-credentials/005-custodian-s3-credentials.md
@@ -57,3 +57,8 @@ so that the custodian can detect and authenticate the user.
 - It would be nice if the user could get some kind of feedback from this process.
 As it stands right now the user should just see the session "magically" appear in
 the renku UI after they click the link.
+- Do we need to support anonymous sessions? The assumptions so far has been no and
+I think it is best that it stays that way so that we keep things simple. If we wanted
+to support sessions started by users that are registered with the Custodian
+but not with Renku then things become a bit more complicated. Namely the Custodian
+would have to follow different flows for registered vs anonymous sessions.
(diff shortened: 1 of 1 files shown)


rokroskar:
@martinfontanet could you please weigh in here on the RFC? If it looks good and is workable we should finalize it and merge.

martinfontanet:
I just added some changes to the diagram in #10.

**Regarding the first question**:
At the moment, the Custodian uses the Renku's Keycloak instance as IDP. It uses OpenID connect (from Renku Keycloak) to authenticate users at both the contract signature step and the session creation step, so it does not need any extra credentials. This means that the link only requires the necessary parameters to start a session with the Renku API (except the S3 credentials, which are handled by the Custodian).

This also partly answers the third question, as there are currently no users that are registered with the Custodian but not with Renku.

rokroskar:
@olevski @martinfontanet what's the status here? would be nice to merge it if it's being used/implemented/considered, otherwise lets close it?

olevski:
@rokroskar @martinfontanet has done some work on this as far as I know. Not sure where the code is or what the status is.

olevski:
@rokroskar this is now up-to-date and ready to be merged.

martinfontanet (approved):

design/005-custodian-s3-credentials/005-custodian-s3-credentials.md
```diff
@@ -0,0 +1,65 @@
+- Start Date: 2022-04-28
+- Status: Draft
```
rokroskar, line 2: I presume this is now beyond a draft? e.g. "partially implemented"?

rokroskar (changes requested):
Please update the status

design/005-custodian-s3-credentials/005-custodian-s3-credentials.md
olevski: Yeah I will change that

rokroskar (approved):

PR #12, opened 2022-07-22
feat: add RFC about new metadata fields for datasets

cmdoret:
Proposal to improve or add new metadata fields for datasets to make them more FAIR.

### DIFF
--- a/design/010-new-dataset-fields/010-new-dataset-fields.md
+++ b/design/010-new-dataset-fields/010-new-dataset-fields.md
@@ -49,8 +49,21 @@ Data usage license. The license cannot live _only_ as metadata and there must al
 * **Mapping:** [license](https://schema.org/license)
 * **Value:** [URL](https://schema.org/URL) from https://spdx.org/licenses
 * **Example:**
+
 ```json
 "https://schema.org/license" : "https://spdx.org/licenses/MIT.html"
+```
+* **Alternative with DigitalDocument**:
+```json
+{
+  "@context": "http://schema.org/",
+  "@type": "DigitalDocument",
+  "name": "MIT License",
+  "identifier": "MIT",
+  "url": "https://spdx.org/licenses/MIT.html",
+  "author": "Alice",
+  "text": "[license text]"
+}
 ```
 
 > Note: See [here](https://github.com/spdx/license-list-data/blob/master/accessingLicenses.md) for programmatic access to the list of licenses and the details of each license. If a license already exists in the project, SPDX also provides a [license matcher](https://github.com/spdx/spdx-license-matcher) to infer the license type from the text.
@@ -85,6 +98,7 @@ The research field. Can be a flat list of keywords, or ideally fine grained term
 * **Mapping:** [size](https://schema.org/size)
 * **Value:** [Text](https://schema.org/Text)
 * **Example:**
+
 ```json
 "https://schema.org/size": {
     "@type": "QuantitativeValue",
--- a/design/010-new-dataset-fields/010-new-dataset-fields.md
+++ b/design/010-new-dataset-fields/010-new-dataset-fields.md
@@ -50,6 +50,8 @@ Data usage license.
 * **Value:** [URL](https://schema.org/URL) from https://spdx.org/licenses
 * **Example:** `https://schema.org/license: https://spdx.org/licenses/MIT.html`
 
+> Note: See [here](https://github.com/spdx/license-list-data/blob/master/accessingLicenses.md) for programmatic access to the list of licenses and the details of each license.
+
 **Affiliation**
 
 For visibility and citability, affiliations should have a unique identifier. Affiliation already exists as a property of the dataset's creator. Ideally we could add a unique identifier to it. Rather than a property of the Dataset, this may be a property of the creator's `affiliation` and would also be available for projects.
--- a/design/010-new-dataset-fields/010-new-dataset-fields.md
+++ b/design/010-new-dataset-fields/010-new-dataset-fields.md
@@ -103,9 +103,16 @@ The research field. Can be a flat list of keywords, or ideally fine grained term
 * **Reason:** user feedback
 * **Mapping:** [size](https://schema.org/size)
 * **Value:** [Text](https://schema.org/Text)
-* **Example:** `https://schema.org/size: "130MB"`
+* **Example:**
+```json
+"https://schema.org/size": {
+    "@type": "QuantitativeValue",
+    "value": 130,
+    "unitCode": "AD"
+}
+```
 
-> Note: It may not be possible to set it for external datasets.
+> Note: It may not be possible to set it for external datasets. Also, [contentSize](https://schema.org/contentSize) could be more appropriate, but for now it only supports text.
 
 ## Drawbacks
 
--- a/design/010-new-dataset-fields/010-new-dataset-fields.md
+++ b/design/010-new-dataset-fields/010-new-dataset-fields.md
@@ -43,14 +43,17 @@ The following fields can be identified from the data above. They are listed in o
 
 **License**
 
-Data usage license.
+Data usage license. The license cannot live _only_ as metadata and there must also be a file in the project. Therefore, this field should be populated when the file is added. In the future, Renku could propose a (graphical/no code) way to choose a license, which would both populate the metadata field and add the file. Here, we simply propose to make this optional field available for future use.
 
 * **Reason:** FAIR compliance
 * **Mapping:** [license](https://schema.org/license)
 * **Value:** [URL](https://schema.org/URL) from https://spdx.org/licenses
-* **Example:** `https://schema.org/license: https://spdx.org/licenses/MIT.html`
+* **Example:**
+```json
+"https://schema.org/license" : "https://spdx.org/licenses/MIT.html"
+```
 
-> Note: See [here](https://github.com/spdx/license-list-data/blob/master/accessingLicenses.md) for programmatic access to the list of licenses and the details of each license.
+> Note: See [here](https://github.com/spdx/license-list-data/blob/master/accessingLicenses.md) for programmatic access to the list of licenses and the details of each license. If a license already exists in the project, SPDX also provides a [license matcher](https://github.com/spdx/spdx-license-matcher) to infer the license type from the text.
 
 **Affiliation**
 
--- a/design/010-new-dataset-fields/010-new-dataset-fields.md
+++ b/design/010-new-dataset-fields/010-new-dataset-fields.md
@@ -64,7 +64,7 @@ For visibility and citability, affiliations should have a unique identifier. Aff
     "@type": "Organization",
     "name": "EPFL",
     "identifier": "https://ror.org/02s376052"
-    }
+    }]
 ```
 
 > Note: ROR has a public REST API which could be used to autosuggest records in a text form in the future. More info [here](https://ror.readme.io/docs/create-affiliation-selection-dropdowntypeahead-widgets)
--- a/design/010-new-dataset-fields/010-new-dataset-fields.md
+++ b/design/010-new-dataset-fields/010-new-dataset-fields.md
@@ -137,6 +137,8 @@ Should some of these fields be editable directly by users on the command line ? 
 
 Would some of these changes introduce incompatibility in (ex/im)porting datasets to/from third party providers ?
 
+How do we deal with missing data ?
+
 Is DSEO's `Domain` class a desirable solution for research topic ?
 
 > What related issues do you consider out of scope for this RFC that could be addressed in the future independently of the solution that comes out of this RFC?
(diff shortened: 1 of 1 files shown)


design/010-new-dataset-fields/010-new-dataset-fields.md
```diff
@@ -0,0 +1,130 @@
...
+
+### User research data
+
+We can leverage information from 3 sources to identify potential fields. In order of decreasing priority: FAIR recommendations, User feedback, User persona.
+
+**FAIR recommendations**
+
+* F1: (Meta)data are assigned a globally unique and persistent identifier
+  + _Ideally we would use something like ROR for **affiliation**_
+* R1.1: data are released with a clear and accessible data usage **license**
```
Panaetius, line 26: :+1:

design/010-new-dataset-fields/010-new-dataset-fields.md
```diff
@@ -0,0 +1,130 @@
...
+
+**Feedback**
+
+* I would like to know the total dataset **size** to adjust session ressources and avoid running out of disk space
```
Panaetius, line 30: :+1:

design/010-new-dataset-fields/010-new-dataset-fields.md
```diff
@@ -0,0 +1,130 @@
...
+* Dataset imports are slow
+  + _A **size** attribute may help by showing an ETA or even “large dataset” warning in the future._
+* I would like to search entities by **affiliation** and **topic** in the search page
```
Panaetius, line 33: is topic covered by keywords?

design/010-new-dataset-fields/010-new-dataset-fields.md
```diff
@@ -0,0 +1,130 @@
...
+* **Mapping:** [creator](https://schema.org/creator) > [affiliation](https://schema.org/affiliation) > [Organization](https://schema.org/Organization) > [identifier](https://schema.org/identifier)
+* **Value:** [URL](https://schema.org/URL) from https://ror.org/
+* **Example:**
+```json
+"http://schema.org/affiliation": [
+    {
+    "@type": "Organization",
+    "name": "EPFL",
+    "identifier": "https://ror.org/02s376052"
+    }
```
Panaetius, line 67:
```suggestion
    }]
```
otherwise git preview looks wonky

design/010-new-dataset-fields/010-new-dataset-fields.md
```diff
@@ -0,0 +1,130 @@
...
+Data usage license.
+
+* **Reason:** FAIR compliance
+* **Mapping:** [license](https://schema.org/license)
+* **Value:** [URL](https://schema.org/URL) from https://spdx.org/licenses
+* **Example:** `https://schema.org/license: https://spdx.org/licenses/MIT.html`
+
+**Affiliation**
+
+For visibility and citability, affiliations should have a unique identifier. Affiliation already exists as a property of the dataset's creator. Ideally we could add a unique identifier to it. Rather than a property of the Dataset, this may be a property of the creator's `affiliation` and would also be available for projects.
```
Panaetius, line 55: It sounds like we'd want a new global node type that can be shared in many projects/datasets. That might be a bit larger effort.

The main downside I see of ROR (or at least its REST API) is that it would add internet connectivity as a requirement to some dataset commands that currently don't have that. And it might be down or have other issues like a schema update, which could then break functionality on our end.
Having it as an extra, optional property would be fine, alongside a unique ID generated by us. Though that somewhat defeats the purpose.

design/010-new-dataset-fields/010-new-dataset-fields.md
```diff
@@ -0,0 +1,130 @@
...
+
+**Topic**
+
+The research field. Ideally, fine grained terms with hierarchical relationships. This would also be relevant for projects.
+
+* **Reason:** User persona + user feedback
+* **Mapping:** [about](https://schema.org/about)
+* **Value:** Maybe [DefinedTerm](https://schema.org/DefinedTerm)
+
+> Note: I am not sure what's the best way of implementing this. [Dataverse](https://demo.dataverse.org/dataverse/demo/search) uses a (short) flat list of terms, whereas [Swiss UBase](https://www.swissubase.ch/en/) uses a fine grained nested list. The closest thing I found is the ["Data Science Education Ontology (DSEO)"](https://fairsharing.org/FAIRsharing.7p0xdg). It has a [Domain](https://bioportal.bioontology.org/ontologies/DSEO/?p=classes&conceptid=http%3A%2F%2Fbigdatau.org%2Fdseo%23domain) class which subclasses owl#Thing.
```
Panaetius, line 80: As mentioned above, `keywords` would cover this at dataverse's level  of detail.

design/010-new-dataset-fields/010-new-dataset-fields.md
```diff
@@ -0,0 +1,130 @@
...
+
+**Size**
+
+* **Reason:** user feedback
+* **Mapping:** [size](https://schema.org/size)
+* **Value:** [Text](https://schema.org/Text)
```
Panaetius, line 86: I'm not sure about text. Having either a custom rdf datatype or using some unit of measure ontology that specifies the unit makes more sense to me.

But then of course using a custom ontology would mean breaking the schema.org ontology (where the `size` term is more about physical product size e.g. centimeters anyways...). Which is one of the more annoying things in RDF.

So like `https://schema.org/size: "130"^^mb` maybe?
And probably we should just keep it in bytes and not do a conversion/rounding, leaving that up to the client/reader?

design/010-new-dataset-fields/010-new-dataset-fields.md
```diff
@@ -0,0 +1,130 @@
...
+## Drawbacks
+
+> Why should we *not* do this? Please consider the impact on users,
+on the integration of this change with other existing and planned features etc.
+
+Proliferation of unnecessary fields may impact the performance and size of the knowledge graph, thus negatively affect user-experience.
+
+> There are tradeoffs to choosing any path, please attempt to identify them here.
+
+This may introduce problems with backwards compatibility, since the metadata structure would be modified.
```
Panaetius, line 100: I think it's more of a problem of how to deal with missing values, which we should consider.

design/010-new-dataset-fields/010-new-dataset-fields.md
```diff
@@ -0,0 +1,130 @@
...
+
+> Why is this design the best in the space of possible designs?
+
+Adding generic information directly in the dataset metadata will make it easier for users to provide crucial context on their data. It is a first step towards encouraging them to fill in these fields.
+
+Currently, affiliation is a free-text value, which means that different values can refer to the same institution (e.g. UNIL and University of Lausanne), or a value may be ambiguous (UCD for University College Dublin or University of California, Davis).
+
+> What other designs have been considered and what is the rationale for not choosing them?
+
+Another option would be to keep these fields as custom annotations and provide tools to help users with data entry. The main issue with this solution is that the fields will not be available in the global KG and thus cannot be used in the search (e.g. to filter datasets in "Machine learning").
```
Panaetius, line 112: Global KG should really get its act together and deal with custom metadata :stuck_out_tongue_closed_eyes:

design/010-new-dataset-fields/010-new-dataset-fields.md
```diff
@@ -0,0 +1,130 @@
...
+The following fields can be identified from the data above. They are listed in order of priority with the following logic: FAIR recommendation > user feedback > user persona.
+
+**License**
+
+Data usage license.
+
+* **Reason:** FAIR compliance
+* **Mapping:** [license](https://schema.org/license)
+* **Value:** [URL](https://schema.org/URL) from https://spdx.org/licenses
+* **Example:** `https://schema.org/license: https://spdx.org/licenses/MIT.html`
```
Panaetius, line 51: this actually raises a legal question. Many of the licenses require to be stored as a file on disk alongside the other files. Having a property in metadata somewhere is ok to show this information, but probably not enough to be in compliance with the license. So I wonder if we should add some support for that as well?

design/010-new-dataset-fields/010-new-dataset-fields.md
cmdoret: That could work! This would require a special checkbox menu in the UI to select "topic" keywords among a set of possible values (like in dataverse). Alternatively, using autocomplete suggestions like for repository topics on github.

design/010-new-dataset-fields/010-new-dataset-fields.md
cmdoret: I agree that affiliation should remain optional. Would it make sense to have "identifier" (ROR id) as an optional property of the organization to avoid making internet connectivity a hard dependency. Maybe with something like this:
1. User adds creator with an affiliation using `renku {project,dataset} {create,edit}`
2. renku-python tries to send query to ror.org
  + :x: (no internet connection or server down) -> identifier is not added to the organization and Renku warns the user, e.g.: :warning: Renku could not connect to the internet to associate a unique identifier with [affiliation].
  + :heavy_check_mark:  -> the ror ID is added as "identifier" to the organization

design/010-new-dataset-fields/010-new-dataset-fields.md
cmdoret: A middle ground betweeen using Text and a custom ontology may be to use QuantitativeValue with a [unitCode](https://schema.org/unitCode) (Here AD is the UN/CEFACT for byte). Is this overkill ?
```json
"https://schema.org/size": {
    "@type": "QuantitativeValue",
    "value": 130,
    "unitCode": "AD"
}
```

design/010-new-dataset-fields/010-new-dataset-fields.md
Panaetius: The other option (or maybe in addition to the above?) would be to try and fill this in when generating triples in the KG using rp. We can assume to have an internet connection there (though ROR might still be down).

cmdoret: We likely need a user action to disambiguate ROR results. Given an input string, the ROR API may return multiple affiliations and it would be risky to assume that the first result is correct. On the command line, the user would at least be able to select among a few options (like for `renku rollback`).

Is there currently not a global user/person node type ? The best would probably be to interactively set the affiliation once (maybe in the UI) and have the ROR id (and ORCID) associated with them in the KG. Then instead of querying ROR for every project or dataset, the user metadata could just be fetched from there?

design/010-new-dataset-fields/010-new-dataset-fields.md
```diff
@@ -0,0 +1,158 @@
...
+* No visibility on my project
+  + _Unique **affiliation** (ROR, ORCID) + **topic** may help with that_
+
+### Proposed fields
+
+The following fields can be identified from the data above. They are listed in order of priority with the following logic: FAIR recommendation > user feedback > user persona.
+
+**License**
+
+Data usage license. The license cannot live _only_ as metadata and there must also be a file in the project. Therefore, this field should be populated when the file is added. In the future, Renku could propose a (graphical/no code) way to choose a license, which would both populate the metadata field and add the file. Here, we simply propose to make this optional field available for future use.
```
mohammad-sdsc, line 46: Datasets already have a `license` field: https://github.com/SwissDataScienceCenter/renku-python/blob/fb76eadb98727589fee51f2d4df768f475a704c1/renku/command/schema/dataset.py#L163. It's populated when importing a dataset. ATM, it's not possible to set/edit this field via CLI/UI. Moreover, it only accepts a URL but according to schema.org (https://schema.org/license) it should be possible to have a `schema:CreativeWork` that should address the mentioned file use-case.

design/010-new-dataset-fields/010-new-dataset-fields.md
rokroskar: Affiliation should be a property of the user in Renku. It shouldn't need to be added on every action or every time triples are generated. I think it's a bit out of scope for the proposal here, but definitely something we need to do. I like the ROR idea (minus the caveats mentioned by @Panaetius) - lets keep it for a proposal on how to properly manage user metadata.

design/010-new-dataset-fields/010-new-dataset-fields.md
cmdoret: How do you think `CreativeWork` would support the file use-case ? Do you mean by allowing to easily create a file from the CreativeWork element ?

If so, SPDX also provides JSON with the license contents for each license URL, e.g. https://spdx.org/licenses/BSD-3-Clause.json

Do you think that would work ?

mohammad-sdsc: I guess either we can have a file as mentioned in the use-case by using `schema:DigitalDocument` which is derivative from `schema:CreativeWork` or we can include the license text in the metadata directly (and not having a file) by using https://schema.org/text.

cmdoret: I added the alternative solution with DigitalDocument to the document !

Panaetius (approved):

mohammad-sdsc (approved):
Thank you!

PR #15, opened 2022-09-15
[Pitch] Friendlier Workflows: adding a Workflow Definition File

lokijuhy:


rokroskar:
I find the title a bit difficult to understand (I know what you mean by it, but if I think about reading it as an "outsider" it sounds strange). How about "Simplified declarative workflows" or something like that?

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
```diff
@@ -0,0 +1,109 @@
...
+Currently, the user has the option of creating a file specifying workflow parameters, and passing this in to `renku workflow execute` to override the default parameters. In the future, the combination of this parameter file and the workflow definition file could develop into a sort of base workflow + several “variation” files defining a set of related workflow experiments. But work into this direction is _out of scope_ of this pitch. 
+
+
+**Should one workflow file be able to reference another workflow file? (i.e. to compose workflows between files)**
+
+My default is *no*, keep it simple for the scope of this pitch.
+
+**Potential for Un-executable `Plans`**
+
+In the current `renku run` implementation, a `plan` is only created after the command executed successfully. By importing and creating the `plan` upon executing the workflow file, the user could potentially end up with a non-executable `plan` in their project. This sounds non-ideal, however, it doesn’t really have any implications for the user, so it’s up to the team for how they would like to handle this behavior.
```
rokroskar, line 101: The related problem I see here is the issue of _when_ one of the workflow files imagined in this pitch gets processed, i.e. added to the metadata of the project. Does it get processed when the file is committed via some sort of hook? Or does it not exist in the project's metadata until it is used? Both options are not ideal - hooks are inherently imperfect (the user simply might not have them installed) and if we wait for the workflow to actually be used before it "exists", we automatically invalidate any "worklow-in-progress" concept that might be natural to have e.g. in the UI.

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
```diff
@@ -0,0 +1,109 @@
...
+
+## 🙅‍♀️ No-gos
+
+There are a lot of feature-rich workflow tools out there already. We’re not trying to make the *best* workflow tool, and not even a comprehensive one. If we can get a researcher started using workflows for the first time because simple workflows are easy to create in Renku, and then that user switches to a different workflow tool that suits them better, that’s a win for Open Research!
+
+We’re not trying to make a comprehensive workflow tool with lots of functionality. We’re trying to make the *minimal* workflow tool a user can learn, so they can take their research project to the next level of reproducibility. So, the file-based workflow syntax need only support a minimal set of workflow functionality.
+
+Specifically, expressing `renku workflow iterate` in the workflow file definition syntax is out of scope, along with any kinds of looping, branching, and wildcard syntax.
```
rokroskar, line 109: The idea of `iterate` is anyway to take an existing workflow definition and iterate through parameters defined therein - so I see it fully compatible with the feature laid out in this pitch by default.

rokroskar (commented):
This is great @lokijuhy thanks for putting it together! It seems to me that the main unresolved concept in this pitch is to determine _when_ the workflow that is created solely via a file makes it to the project's metadata. 

One tool that has a very similar approach is dvc - there the workflow "template" is basically doing "dvc run" -- i.e. that creates a workflow file that you can edit afterwards, add more steps etc. I think some connection between the workflow plans that are created via `renku run` and the proposal here should be made - do we need a way to e.g. save a plan that was generated automatically to a file for editing?

lokijuhy:
> This is great @lokijuhy thanks for putting it together! It seems to me that the main unresolved concept in this pitch is to determine _when_ the workflow that is created solely via a file makes it to the project's metadata.
> 
> One tool that has a very similar approach is dvc - there the workflow "template" is basically doing "dvc run" -- i.e. that creates a workflow file that you can edit afterwards, add more steps etc. I think some connection between the workflow plans that are created via `renku run` and the proposal here should be made - do we need a way to e.g. save a plan that was generated automatically to a file for editing?

I see that this aspect was not made clear in the Pitch. I am picturing that the workflow is committed to the Renku project when the user runs it. I don't see a good way of adding it the project (i.e. showing it in the workflow viewer) any earlier than that, and I think that's ok from the user point of view. I added this to the Solution section under "When is the workflow 'saved' to the project?"

As for _generating_ a workflow file from an existing workflow/renku run, I would put that in out-of-scope for now, and see if we get requests from users for that functionality. I added this to the Out of Scope section.

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
```diff
@@ -0,0 +1,109 @@
...
+
+### Editing workflow files
+
+The workflow file is opened and modified just like any other code file in the project. This pitch requires no changes to the UI.
+
+The user should be able to create more than one workflow file in a project.
+
+### Running workflow files
+
+The user would still use the CLI to run the workflow, referencing the workflow via the workflow file name. To improve usability for non-CLI users, in tutorials, we could perhaps demonstrate running shell commands from inside a notebook (`!renku workflow execute workflow.yml`).
```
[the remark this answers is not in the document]
lokijuhy: I see that it makes sense to reference the workflow name as defined in the workflow file, since non-file-based (CLI) workflows are also referenced by name. But do we also open a small can of worms here, by introducing a need for within-workflow file search? For example, I think it should be a must-have that the user can define more than 1 workflow file within a project. Therefore, when referencing workflow names, we have to check for duplicate workflow names within and between files, and raise error messages appropriately. Perhaps this is a simple enough problem to solve, but just want to surface it before committing to a solution. What do you think?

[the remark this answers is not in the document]
rokroskar: yeah, plus if you see the workflow only getting added to the project metadata _after_ it has been run at least once, it won't be possible to find it by name anyway. I suppose it should be easy enough to check whether the `name` in `renku workflow execute <name>` is a _filename_ or a workflow name that already exists in the project.

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
```diff
@@ -0,0 +1,109 @@
...
+
+Developing a workflow should be as simple to the user as editing the workflow file, executing the workflow file on the CLI, making further modifications to the workflow file, executing the workflow file again, etc. Renku should not interrupt or slow down this development process.
```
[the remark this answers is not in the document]
lokijuhy: Yea, I agree that the need to commit before every renku run/workflow command is a small but constant annoyance to usability. What do you mean by "dry-run" exactly? Would it just check that the workflow file syntax is legitimate? 

One option to consider- though I recognize that this is opening a fairly new direction for Renku- is to have `renku workflow execute` automatically do a commit before executing the workflow, so the user doesn't have to do that extra step every time. I haven't thought through this idea very much... But we could offer something like a `renku workflow execute <workflow> --commit-for-me` flag that defaults `True`, but the user could turn off if they don't want commits happening for them...? Totally not sure about this idea, just throwing it out there. I agree the extra commit step is annoying to users. I've even seen it stumble up Renku team members when they're just trying to demo something.

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
```diff
@@ -0,0 +1,109 @@
...
+
+The user should be able to create more than one workflow file in a project.
+
+### Running workflow files
+
+The user would still use the CLI to run the workflow, referencing the workflow via the workflow file name. To improve usability for non-CLI users, in tutorials, we could perhaps demonstrate running shell commands from inside a notebook (`!renku workflow execute workflow.yml`).
+
+Developing a workflow should be as simple to the user as editing the workflow file, executing the workflow file on the CLI, making further modifications to the workflow file, executing the workflow file again, etc. Renku should not interrupt or slow down this development process. 
+
+The user should be able to run a specific named workflow step from a file without running the whole file (a la `renku workflow execute workflow.yml step_1`, or even `renku workflow execute step_1` if we can provide searching all workflow files). As a _nice-to-have_, it would also be nice to support running a subset of steps, for example `renku workflow execute workflow.yml --from step_name_3 --to step_name_6`.
```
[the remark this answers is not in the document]
lokijuhy: I took out the searching for all step names per feedback from Lorenzo, though your comment above about referencing workflows via name vs file name reopens the issue of disambiguating step names...

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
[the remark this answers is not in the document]
rokroskar: right, I think "dry-run" was bad naming - I meant more like just running the workflow as prescribed, but do not commit anything. There was something like this already being done because we need to delay committing in the case that the execution is happening remotely. The idea we had in the past was something like a renku "transaction" - you could do a bunch of work using the Renku CLI and Renku would keep track of the metadata somewhere, but wouldn't commit anything until you told it to.

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
```diff
@@ -0,0 +1,109 @@
...
+
+### Improving workflow user awareness: A workflow file template
+
+It is worth considering adding a template workflow file to the default Renku project structure. For example, this could be a workflow file with an example workflow step written out but commented out. The user can uncomment it and replace the placeholders with their command, inputs, output, etc. There could also be a comment at the top of the file describing how to run the workflow.
```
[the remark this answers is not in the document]
lokijuhy: I don't quite understand what the problem you're picturing here is. Do you mean the issue of the file being named `workflow-template.yaml`? Why don't we just name it `workflow.yaml`? I don't see any problem with that? If they don't use it, it's fine. It's like having an empty `environment.yaml` in the project. It they do, `workflow.yaml` is a fairly good name. It's only if they create a second workflow file that they should name it better, which they hopefully will do...?

I am concerned that if we go the command-to-generate-a-template route, that will have _substantially_ less impact towards putting workflows in front of users.

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
```diff
@@ -0,0 +1,109 @@
...
+
+We’d like to offer the ability to define Renku workflows in a file, where the full workflow is laid out for the user to plainly see and edit.
+
+I’ve done a [survey of popular workflow tools](https://sdsc.atlassian.net/wiki/spaces/RENKU/pages/2310701105/Workflow+Tool+Research), and YAML seems to be a common choice for a workflow file, but we leave the final decision to the team.
+
+<img src="fat-marker-file.png" width="400">
+
+The workflow file names a sequence of steps, and each step names its command, inputs, outputs, parameters, etc. Each step should be named so that it can be referenced for running individually. I imagine that the structure of the workflow step definition in the file would be pretty similar to the output of `renku workflow show`.
+
+![renku workflow show](workflow-show.png)
```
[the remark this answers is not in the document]
lokijuhy: Parameters is already listed there, or am I missing something about your comment?

I will add annotations & metadata.

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
```diff
@@ -0,0 +1,109 @@
...
+
+
+**How do we meaningfully list *relevant* workflow metadata with so many pieces of workflow metadata floating around?**
+
+The one cost of creating lots of workflow metadata is that listing the project's workflows yields a lot of metadata that the user doesn't care about. But the solution to that is right in the statement- the user doesn't care about any metadata that relate to data objects that no longer exist. `renku workflow ls` should only list workflows that relate to currently existing code and data objects, i.e. show me the workflow that I ran to create the current version of file `b`, not the workflow I ran yesterday that created a previous version of file `b`. (In the `plan` and `activity` vocabulary, only `plans` that have an `activity` that relates to a file currently in the project)
```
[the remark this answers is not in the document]
lokijuhy: Could you elaborate on what you mean?

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
```diff
@@ -0,0 +1,109 @@
...
+
+This would put workflows in front of more users, increasing awareness and hopefully also adoption. 
+
+### Note: Re-runnability & Pre-existing Outputs
+
+A common frustration users have with current Renku workflows is that the `workflow` & `run` (mostly `run`) commands errors when you run a workflow command that writes an output file that already exists.
+
+Users do this frequently during development: run script `a`, generate file `b`. Change script `a`, rerun, generate a new b. Great, now you're happy with script `a`, so `renku run script_a` one last time to record it. But now Renku complains that no outputs were generated (because file `b` already existed, so nothing was detected)! Users *can* avoid this by telling Renku explicitly that the output is file `b`, but that's more typing. More commonly, users end up deleting output files before every `renku run`, which they- very reasonably- don't like doing.
+
+The file-based workflow feature should solve this. The workflow should read the specified output in the workflow file, rather than detecting changed files. This removes the need for the user to delete the file so that Renku can detect its creation.
```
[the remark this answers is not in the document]
lokijuhy: As a piece of counterpoint- I have heard feedback from users that the output detection is annoying when the user creates a log file as part of their workflow- they don't want this tracked as an output.

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
lokijuhy: Could you elaborate on the "workflow-in-progress" concept? I'm not sure what you mean.

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
lokijuhy: From my discussion with Ralf, my understanding is that this is not trivial to implement - @Panaetius ?

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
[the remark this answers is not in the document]
rokroskar: true, I was looking at the sketch 🙈

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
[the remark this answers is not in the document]
rokroskar: true! Maybe there should be some way of turning it off/on completely?

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
rokroskar: I think it's moot given that somewhere else it was decided that the workflow needs to be executed before it gets added to the project's metadata. Until that point it's basically invisible in any workflow front-end.

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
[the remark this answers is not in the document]
lokijuhy: This feels like it's opening a whole separate topic that could be it's own pitch...?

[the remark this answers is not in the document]
rokroskar: I think it will likely come up during design/testing/ux validation so maybe worth adding it to the out-of-scope section.

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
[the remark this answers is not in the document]
lokijuhy: I'll expand the pitch to say via the workflow _via the workflow file name **or** the workflow name (TBD exactly how this works)_, and leave this up to the implementation team.

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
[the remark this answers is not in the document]
rokroskar: hmm that's true. I guess there's no harm in a default `workflow.yaml`. 👍

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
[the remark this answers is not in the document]
lokijuhy: Added.

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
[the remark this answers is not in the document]
lokijuhy: And by _retrieving_ , do you mean `renku workflow show`? `renku workflow execute`? Pulling up the previous version of the workflow definition file?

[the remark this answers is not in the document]
rokroskar: yes, finding out which version of the workflow created which version of an output. I suppose git would tell you that anyway, but I didn't know if you envisioned some way to step through the different versions. That can also be labelled as out-of-scope I suppose.

lokijuhy:
> I find the title a bit difficult to understand (I know what you mean by it, but if I think about reading it as an "outsider" it sounds strange). How about "Simplified declarative workflows" or something like that?

I changed it... 👍 👎 ❓

Panaetius (commented):
I'm a bit worried about deleting Plans and creating new ones with the same name on each `wf execute`, it pollutes the KG and we can't distinguish between a workflow a user deleted because they thought it's garbage and one we deleted on their behalf because of this feature.

Executing workflow files directly is natural to users, so I understand the motivation for this pitch. But having `renku workflow execute` import and then execute a workflow, with the import logic being somewhat hidden from users, does feel like a bit of a hack on top of the existing system and not the cleanest solution. Renku is a bit unique in that our workflow definition is KG based, not file based, motivated in part to support exploring/analyzing workflows from the KG directly and "doing unspecified smart things" :tm: with it at some point in the future. And this pitch is essentially a (little bit clunky) way of mirroring changes to a file into the KG. It solves the "run a workflow file" problem at the cost at making the KG less correct/representative. And there's also some edge-cases where the two can collide, e.g. if have a run with some name already, and then create a new file to create a composite and name one of the steps the same as the existing run (even though the two are unrelated). We'd either have to fail on every name conflict (not feasible as then this feature would only work once, for each workflow file, and then yield name conflicts) or potentially delete and overwrite a workflow that's perfectly fine and valid and should not be deleted. And having logic to detect which things changed and only partially updating things, so at least we don't need to create x new Plans every time a user runs a workflow file, only addresses some parts of this.

So reading through this, I can't help but wonder if it's not going far enough? Instead of trying to squeeze this functionality into the existing concepts in the metadata that don't fit it that well, I wonder if we could expand the metadata to support it better? We could
- think about making workflows properly mutable, so we don't need to delete and create new ones. There's good reasons that this isn't the case right now, and it makes things easier on our end, but maybe it would be possible. And I know I said before that this is tricky to change (especially regarding dependencies between composites and their children), but down the line it might save us work
- Add a new type of workflow (or several?) that can be more directly tied to a file? Like instead of `Plan`, have a `FileBasedPlan` which keeps track of which workflow file it's based on and exposes all the relevant properties in the KG, but is essentially read-only and can only be modified through editing the file, not the CLI. I.e. it'd be more of a view on the file than in the KG, instead of being a disconnected entity. Haven't thought this fully through yet, but could be a nice way of separating things. Plus then someone else could know that Plan X is based on workflow_x.yml.

Or both of the above?

I think in the end, there's 3 separate use-cases that do overlap but are all conflated into one here. Namely:
1. creating Plans from a file because the CLI is too cumbersome for complicated things
2. executing workflows from file instead of through metadata
3. Tracking commands a user's runs as Activity's for reproducibility

and the solution here is to always do all of the above, each time. And I'm not sure that needs to be the case. For instance in dev mode, I only really care about 2. Once I'm happy with my code, I mostly care about 3 and 1 or maybe 2. And for sharing with others, 1 is important.

Before rambling on, I just feel that since we are free to change our metadata, we might be able to change/extend it such that this use-case fits more naturally, instead of trying to make the use-case fit with what is there.

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
```diff
@@ -0,0 +1,122 @@
...
+### When is the workflow "saved" to the project?
+
+The workflow defined in a workflow file is "saved" to the Renku project when the user executes the workflow. This means the workflow defined in the workflow file is not in the project metadata before the user runs it.
+
+### Editing workflow files
+
+The workflow file is opened and modified just like any other code file in the project. This pitch requires no changes to the UI.
+
+The user should be able to create more than one workflow file in a project.
+
```
mohammad-sdsc, line 44: We should do some basic validation on the workflow files (e.g. there are no loops) when adding and editing workflow files. We can also do the validation when running the workflow file.

rokroskar:
@Panaetius I think you raise some good points - but there are a few things that I'm a little confused about. 

> I'm a bit worried about deleting Plans and creating new ones with the same name on each `wf execute`, it pollutes the KG and we can't distinguish between a workflow a user deleted because they thought it's garbage and one we deleted on their behalf because of this feature.

Where does this need for deletion come from? I can right now create a workflow, give it a name, and edit it. It will retain the name, but e.g. provide different default values - in my history, I still have two workflows (same ID, same name, but different details). Fundamentally, the use-case described here is the same. 

> Executing workflow files directly is natural to users, so I understand the motivation for this pitch. But having `renku workflow execute` import and then execute a workflow, with the import logic being somewhat hidden from users, does feel like a bit of a hack on top of the existing system and not the cleanest solution. Renku is a bit unique in that our workflow definition is KG based, not file based, motivated in part to support exploring/analyzing workflows from the KG directly and "doing unspecified smart things" ™️ with it at some point in the future. And this pitch is essentially a (little bit clunky) way of mirroring changes to a file into the KG. It solves the "run a workflow file" problem at the cost at making the KG less correct/representative. 

I don't fully understand this. I see the workflow file approach essentially the same as the `renku run` approach - they both translate the intent of the user into an abstract KG representation of a series of steps. In one case, it's based on what the user actually ran, in the other case it's what they would like to run. So maybe this needs to be a different `renku run` type command rather than `renku workflow execute`? Would that make it cleaner?

> So reading through this, I can't help but wonder if it's not going far enough? Instead of trying to squeeze this functionality into the existing concepts in the metadata that don't fit it that well, I wonder if we could expand the metadata to support it better? We could
> 
>     * think about making workflows properly mutable, so we don't need to delete and create new ones. There's good reasons that this isn't the case right now, and it makes things easier on our end, but maybe it would be possible. And I know I said before that this is tricky to change (especially regarding dependencies between composites and their children), but down the line it might save us work

They seem to be mutable already? Or are they only partially mutable?
 
>     * Add a new type of workflow (or several?) that can be more directly tied to a file? Like instead of `Plan`, have a `FileBasedPlan` which keeps track of which workflow file it's based on and exposes all the relevant properties in the KG, but is essentially read-only and can only be modified through editing the file, not the CLI. I.e. it'd be more of a view on the file than in the KG, instead of being a disconnected entity. Haven't thought this fully through yet, but could be a nice way of separating things. Plus then someone else could know that Plan X is based on workflow_x.yml.

I think this makes a lot of sense tbh - but seems to require some sort of git hook functionality? Or I guess the dependency of a `FileBasedPlan` is on the file checksum that we could check any time the workflow is invoked...?

> Before rambling on, I just feel that since we are free to change our metadata, we might be able to change/extend it such that this use-case fits more naturally, instead of trying to make the use-case fit with what is there.

That's potentially a great avenue, but I see that as an implementation decision which is not the concern of this document.  I think the intent is valid - specify in some simple declarative way some steps of a workflow and the goal of this doc should be to limit the scope of what that means/implies. IMHO even the commands are free to change or be re-interpreted as the implementation takes place as long as the higher-level goals are served.

Panaetius:
@rokroskar 
> Where does this need for deletion come from? I can right now create a workflow, give it a name, and edit it. It will retain the name, but e.g. provide different default values - in my history, I still have two workflows (same ID, same name, but different details). Fundamentally, the use-case described here is the same.

yes, but we limit that to essentially non-functional changes, you can change the name, description, etc. but not the structure (add a new parameter). The only way to do that currently would be to delete the plan and create a new, unrelated one with the same name. Which is what the `What happens when you modify a workflow file?` section describes, because that's how it currently works and what I told Laura way back when, but now I'm not sure it actually is a feasible solution.

> I don't fully understand this. I see the workflow file approach essentially the same as the renku run approach - they both translate the intent of the user into an abstract KG representation of a series of steps. In one case, it's based on what the user actually ran, in the other case it's what they would like to run. So maybe this needs to be a different renku run type command rather than renku workflow execute? Would that make it cleaner?

Yes that'd make sense I think, like `renku run -f workflow.yml`, then the semantics are consistent (renku wf execute executes something already in the KG, renku run executes a new thing and creates a new plan).

> They seem to be mutable already? Or are they only partially mutable?
Only name, description, keywords, creators (soon) and default value can be changed.

> I think this makes a lot of sense tbh - but seems to require some sort of git hook functionality? Or I guess the dependency of a FileBasedPlan is on the file checksum that we could check any time the workflow is invoked...?

I wouldn't use hooks. It could be on a file checksum with some update command/flag or just be auto-updated on re-execution?

> That's potentially a great avenue, but I see that as an implementation decision which is not the concern of this document. I think the intent is valid - specify in some simple declarative way some steps of a workflow and the goal of this doc should be to limit the scope of what that means/implies. IMHO even the commands are free to change or be re-interpreted as the implementation takes place as long as the higher-level goals are served.

I fully agree, it's more that the current document already talks about what should happen in the metadata to some extent, and that can probably be left out :slightly_smiling_face:

lokijuhy:
@Panaetius 

> So reading through this, I can't help but wonder if it's not going far enough? Instead of trying to squeeze this functionality into the existing concepts in the metadata that don't fit it that well, I wonder if we could expand the metadata to support it better?

> ... it's more that the current document already talks about what should happen in the metadata to some extent, and that can probably be left out

Excellent points! This has already been a learning experience in how to shape a good pitch :smile: I am open to removing the metadata-related Rabbit Holes, and moving the Pitch to a higher level of abstraction. I would leave in all of the user-behavior constraints that I described that we want to satisfy, but not prescribe how to achieve them technically. 

However, the one big question I have as a Shaper is _What questions **do** we need to resolve in advance of starting this pitch to ensure that it does not spiral beyond 6 weeks_? Do you think that if I removed the rabbit hole technical details, the team could resolve the technical implementation in 6 weeks? What rabbit holes _do_ we need to resolve in advance to ensure a smooth delivery?

lokijuhy:
@Panaetius 
> I think in the end, there's 3 separate use-cases that do overlap but are all conflated into one here. Namely:
> 
> 1. creating Plans from a file because the CLI is too cumbersome for complicated things
> 2. executing workflows from file instead of through metadata
> 3. Tracking commands a user's runs as Activity's for reproducibility
> 
> and the solution here is to always do all of the above, each time. And I'm not sure that needs to be the case. For instance in dev mode, I only really care about 2. Once I'm happy with my code, I mostly care about 3 and 1 or maybe 2. And for sharing with others, 1 is important.

I like where you're going, pulling apart the pieces at play here. However, I see it a different way. A big benefit to writing workflows as files is that the workflow file _is_ metadata. The act of writing out the workflow file _is_ tracking your commands for reproducibility. To throw a radically different idea out into the arena: What if file-based workflows didn't create `Plans` at all, or at least no more than a pointer in the KG to the file? (Perhaps this aligns with your `FileBasedWorkflow` idea?) The workflow file has got you covered for recording the workflow metadata itself.

Second, I disagree that there are clean divisions between "dev mode" and "final run mode". In my experience, it's always "dev mode", and at some point you decide what you've got is good enough. But a user rarely knows in advance whether _this run_ is _the run_. And, a user is rarely willing to re-run their "final workflow" in a separate step for the sake of tracking it (unless running the workflow is _very fast_, which is uncommon in data science). So, we need to support a development flow that enables effortless edit-ability (2) and tracking (3) at the same time, all the time. It may be worthwhile to break these pieces apart behind the scenes in the technical implementation, but we need to keep them together in a single user experience in order for users to adopt our system.

Panaetius:
> (Perhaps this aligns with your FileBasedWorkflow idea?
Yes I think it goes in a similar direction. The main difference is that I would still like it to go beyond a simple pointer, in that it is more of a view on the file (in the Database view sense of the word "view"). I.e. it still exposes the properties in the knowledge graph, but it can't be edited independent of the file and is 1:1 tied to the file. Only to make it still searchable/analyzable in the KG without having to read the file for this purpose. But it'd be simply a representation of the file in the KG, not a copy/separate entity. I.e. `renku workflow show` works on it but `renku workflow edit` doesn't, to give a concrete example.

As for the second point, I agree, I was more thinking in terms of something we considered in the past: to have a sort of "soft" tracking of executions, in that we would write metadata in a sort of staging area, ephemerally, and not commit it to the KG immediately, but then if you decided "Yep, this is the one!", you could tell renku "ok, track that last one properly". The main concern this would address is to enable a user to lower the signal-to-noise ratio and not have tons of executions recorded and show up that they don't care about.

lokijuhy:
@Panaetius I updated the pitch to reflect our discussion yesterday. Could you take a look?

FYI @rokroskar and @mohammad-sdsc  - Ralf and I discussed all-things-metadata yesterday and agreed to remove much of the implementation details from the Rabbit Holes, and leave only a few high-level metadata requirements.

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
```diff
@@ -0,0 +1,106 @@
...
+<img src="fat-marker-file.png" width="400">
+
+The workflow file names a sequence of steps, and each step names its command, inputs, outputs, parameters, etc. Each step should be named so that it can be referenced for running individually. I imagine that the structure of the workflow step definition in the file would be pretty similar to the output of `renku workflow show`.
+
+As a nice-to-have, the workflow definition file should allow for the specification of workflow annotations and additional metadata.
+
+![renku workflow show](workflow-show.png)
+
+For the scope of this pitch, expressing `renku workflow iterate` in the workflow file definition syntax is _out of scope_, along with any kinds of looping or branching.
```
rokroskar, line 33: That `iterate` is out of scope is almost by-definition - `renku workflow iterate` is to execute an existing workflow with a series of parameters, it's not something expressed in the workflow itself.

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
```diff
@@ -0,0 +1,106 @@
...
+
+## 🐰 Rabbit Holes
+
+### Venturing into the Metadata Realm
+
+This pitch will require a substantial effort to make the metadata all play nice. Without diving into implementation details, here are a few high-level requirments that the metadata must meet in order to provide a good user experience:
+
+* The workflows defined in files should be fully represented in the Renku KG metadata.
+* However, the workflow file itself should be the single source of truth for the workflow definition (not the KG metadata).
+* File based workflows can only be edited via the file. These workflows cannot be edited via the CLI or (in the future) somewhere in the RenkuLab UI.
```
rokroskar, line 72: why not in the UI? That could also just modify the file under the hood.

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
```diff
@@ -0,0 +1,106 @@
...
+* In terms of versioning a file-based workflow, the metadata should contain metadata about the file-based workflow at each execution. There is no need to maintain a history of the file-based workflow at commits where the workflow was modified but not executed.
```
rokroskar, line 73: This first sentence is confusing. You mean to say that the renku KG should capture the file-based WF at execution _only_? I.e. it should not automagically interpret changes to the file via a hook mechanism?

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
```diff
@@ -0,0 +1,106 @@
...
+* However, the KG must be aware of the latest state of the workflow file (even un-executed or not committed), because other parts of RenkuLab will be pulling information about the workflow via the KG (for example, the workflow browser will query the KG for the file-based workflow definition, and this needs to be up to date with the current state of the file so as not to confuse the user, even if the workflow file has not been executed since it was last modified).
```
rokroskar, line 74: I guess this means that something like `renku update` should notice if a file-based workflow has been changed since last execution and offer to run it again?

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
```diff
@@ -0,0 +1,106 @@
...
+
+### Misc Rabbit Holes
+
+**How do you handle when a workflow step generates many files?**
+
+As is the state of the current workflow functionality, inputs and outputs may be directories, rather than files, to support multiple files. However we don’t support any kind of wildcard syntax.
+
+**How does this feature mesh with the current workflow parameter file functionality?**
+
+Currently, the user has the option of creating a file specifying workflow parameters, and passing this in to `renku workflow execute` to override the default parameters. In the future, the combination of this parameter file and the workflow definition file could develop into a sort of base workflow + several “variation” files defining a set of related workflow experiments. But work into this direction is _out of scope_ of this pitch.
```
rokroskar, line 86: I don't really think it needs to be addressed - `renku workflow execute` will continue to work even for file-based workflows (I assume) because once they are in the renku KG they will be accessible for re-execution.

rokroskar (commented):
This is looking good @lokijuhy! Just a few clarifications from my side, but otherwise it seems gtg.

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
Panaetius: Yes the file can be edited through the UI as well, I think, but not the metadata directly, in isolation.

Something like `The metadata of file-based workflows can only be edited simultaneously with the file, not separately`

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
Panaetius: This one comes from me. We probably don't want to
- have a githook to pick up changes and create new plans
- require the user to run a command to update based on an edit 
- walk git history to create plans on the fly for all commits to the file on graph generation (the thought alone of going back to that gives me nightmares).

So the compromise with good UX and technical feasibility is to create a plan object in the repo metadata on execution (if one doesn't exist already) and to also create one on the fly for any modifications that weren't executed, but only for the latest version (no walking of git history or anything). This latest version would then be mutable whereas the rest isn't. 
This is so users don't have to run a special command to update the KG metadata but to also see their most recent edits in the UI (not having them show up would be confusing).

We could leave it open for a hook based approach, but then I don't like adding crucial functionality to the hooks, since they might not be set up and a user might skip them for other reasons (like `pre-commit` failing for a non-issue).

We could still have a command to update metadata from file without execution like `renku workflow update` or something, but that's more of a detail

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
Panaetius: That's a good idea!

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
Panaetius: yes, that's the idea. and they can be used like normal Plans.

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
```diff
@@ -67,24 +63,16 @@ The file-based workflow feature should solve this. The workflow should read the
...
+* However, the KG must be aware of the latest state of the workflow file (even un-executed or not committed), because other parts of RenkuLab will be pulling information about the workflow via the KG (for example, the workflow browser will query the KG for the file-based workflow definition, and this needs to be up to date with the current state of the file so as not to confuse the user, even if the workflow file has not been executed since it was last modified).
+* It has been considered that there might be a need for a command to force-save the file-based workflow to the Renku metadata without executing it. For example, this might be helpful if the user modifies the description of the workflow, and wants to save this information to the KG without running their whole long-running workflow. However, the idea that there is a "save" beyond "file save" is a ugly area we'd like to avoid forcing upon the user if possible, so let's assume this is not needed. But the implementtion team should keep an eye out for other corner cases that might require such a command.
```
Panaetius, line 75: I think the implementation of such a command would be relatively easy, so it's a trade-off between the added complexity of having more commands and the potential need for such a command by users.

Panaetius:
looks good to me too. 

One thing that we didn't go into much detail last week and that I remember last weekend is CompositePlans, which I think should work the same, just that a CompositePlan from a file should be treated as one unit together with its steps. Steps can be executed individually but when editing the file, you always edit the whole composite with all steps. I think that makes things much easier.

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
lokijuhy: @rokroskar FYI One other thing Ralf and I discussed last week is that it makes more sense conceptually for file-based workflows to be run via `renku run` than `renku workflow execute`. This is because `renku workflow execute` executes existing workflows from metadata, while `renku run` runs a user command. Plus, I think users would be much happier with the shorter syntax, too! I think it can be discussed during Build whether we want to support running the file-based workflow from both commands, or just `renku run`, etc.

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
lokijuhy: I added a small section about `renku update`.

lokijuhy:
@Panaetius 
> One thing that we didn't go into much detail last week and that I remember last weekend is CompositePlans, which I think should work the same, just that a CompositePlan from a file should be treated as one unit together with its steps. Steps can be executed individually but when editing the file, you always edit the whole composite with all steps. I think that makes things much easier.

I'm not sure what you mean. What would you like added to the pitch?

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
rokroskar: yep, that makes sense and it's how I understood it but it took me a while to parse the "meta-meta-metadata" piece. 😆

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
rokroskar: Yes definitely, it's how I understand it too. But you'll have something like this (which I think is nice!):

```
$ cat workflow.yml

name: my-workflow
step1: 
  name: do_stuff
  parameters:
    x: 1
    y: 2
step2: 
  name: do_more_stuff
  ...

$ renku run workflow.yml
# runs the workflow

$ renku workflow ls
<id> my-workflow ...

$ renku workflow execute --set step1.x=10 my-workflow
# runs my-workflow, as defined in workflow.yml and sets x=10 in step1
```

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
rokroskar: thanks! Should probably also mention that `renku status` should maybe have a special section for entire workflows that are out-of-date? Right now it's purely based on inputs/outputs.

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
lokijuhy: Yes, seems like a nice freebie 🤷‍♀️

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
lokijuhy: done

Panaetius:
> @Panaetius
> 
> > One thing that we didn't go into much detail last week and that I remember last weekend is CompositePlans, which I think should work the same, just that a CompositePlan from a file should be treated as one unit together with its steps. Steps can be executed individually but when editing the file, you always edit the whole composite with all steps. I think that makes things much easier.
> 
> I'm not sure what you mean. What would you like added to the pitch?

I think it's fine, I mostly just wanted to say when talking about Plans last week, I mostly considered a single step and not a composite. But then everything written should be applicable to composites in the same way. So for a pitch it's enough, though implementation will likely differ a bit from exactly what I had in mind. Things like does each Plan and Composite track the file and checksum, or just the parent composite and the children depend on it. But it's implementation details.

Should we mention that the metadata boundaries we define are minimum requirements? E.g. only tracking plans that were executed+latest plan to me is a minimum, but if whoever implements it comes up with a smart way to track all commits then that's perfectly fine too (though I doubt that's possible)

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
```diff
@@ -0,0 +1,112 @@
...
+
+* The workflows defined in files should be fully represented in the Renku KG metadata.
+* However, the workflow file itself should be the single source of truth for the workflow definition (not the KG metadata).
+* File based workflows can only be edited via the file. These workflows cannot be edited via the CLI or (in the future) somewhere in the RenkuLab UI.
+* In terms of versioning a file-based workflow, the metadata should contain metadata about the file-based workflow at each execution. There is no need to maintain a history of the file-based workflow at commits where the workflow was modified but not executed.
+* However, the KG must be aware of the latest state of the workflow file (even un-executed or not committed), because other parts of RenkuLab will be pulling information about the workflow via the KG (for example, the workflow browser will query the KG for the file-based workflow definition, and this needs to be up to date with the current state of the file so as not to confuse the user, even if the workflow file has not been executed since it was last modified).
```
mohammad-sdsc, line 80: What you do mean by `not committed` in the first sentence? Is it a typo or does it mean that KG should be updated in real-time?

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
```diff
@@ -0,0 +1,112 @@
...
+### Updating file-based workflows
+
+`renku update <file-based workflow>` should check if the workflow file has been changed since last execution as well as if any of the steps' inputs have changed, and offer to run it again. TBD whether we can be smart and only run the affected subset of the workflow (For example, if a workflow has steps A -> B -> C and step B changed, only run B and C, not A).
+
+Simiarly, `renku status` should check for file-based workflows where the definition file has been modified (or inputs have been modified).
+
+### Improving workflow user awareness: A workflow file template
+
+In order to increase workflow usage, we also need to make users aware of them. To this end, it is worth considering adding a template workflow file to the default Renku project structure. For example, this could be a workflow file with a few example workflow steps written out but commented out. The user can uncomment it and replace the placeholders with their command, inputs, output, etc. There could also be a comment at the top of the file describing how to run the workflow.
+
```
mohammad-sdsc, line 58: I'm not sure if we should do this. I believe advertising this new feature in the release note, discord, ... is enough. We should also include it in Renku tutorials.

mohammad-sdsc: I suggest that we add a command to convert a plan/composite-plan to a workflow file. Since the semantics of the workflow file and plans are almost the same, this is a low-hanging fruit.

rokroskar: @mohammad-sdsc why? It's pretty standard for tools to do this, e.g. helm generates a chart template, vscode generates a functional dev-container template etc. I do agree that converting an existing plan to a file would be nice and would completely make sense - not sure why it was explicitly _excluded_ (see the "out of scope" section) - @lokijuhy?

lokijuhy: I do stand by the workflow template file. I think most users do not read the release notes, but more importantly even if they do, it takes seeing something multiple times before someone will test out something new. And I don't think we will _annoy_ anyone by adding it to the project template, so I don't see any downsides. 

As for the conversion-of-existing-plan-into-worfklow-file being in Out of Scope, I'm content to move it to nice-to-have status.

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
lokijuhy: Not a typo... @Panaetius live-updating the KG between commits was part of the implementation proposal we talked through, right? Or did I misunderstand?

I was walking through an example use case of why live-updating to the KG would be important, but then concluded that it would _not_ be so important, so maybe we could remove this requirement? The example I am thinking about is the workflow browser that is being built right now. If a user is writing a workflow in a session, and then looks at the workflow browser to see their workflow, it could be confusing if the workflow browser showed an out of date workflow. But perhaps the user would understand that the workflow browser won't update until they commit?

rokroskar: That sounds very complicated imho - maybe if we make it clear that the workflow state corresponds to a specific state of the repository it would be enough of a clue for most users?

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
rokroskar: I guess the two could be linked? I.e. I could do 

```
renku workflow edit --description "bla" my-workflow
```

and the command would edit the workflow in the metadata _and_ in the file?

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
Panaetius: I don't think we want live updating with uncommitted stuff. but we want to at least process the latest commit (if not executed) so a user doesn't go "Why don't I see my changes in the UI that I committed and push?"

rokroskar: So "not committed" should be dropped from that sentence.

lokijuhy: Ok, fixed.

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
lokijuhy: Unless this is super easy to implement, I would leave this out of scope, because I don't imagine many users would care about being able to do this. And rather return am error message that file-based workflows can only be edited via the file.

rokroskar: sure that's fine - lets add it to the out-of-scope section

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
rokroskar: the only downside I see to this is that we get a bunch of workflows that look identical. But then that happens anyway if you make it a part of a tutorial, so leaving it out doesn't really avoid the problem.

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
```diff
@@ -0,0 +1,116 @@
+# Broadly Accessible Workflows via a Workflow File
```
rokroskar, line 1:
```suggestion
# Friendlier Workflows: adding a Workflow Definition File
```

lokijuhy:
@rokroskar @Panaetius @mohammad-sdsc Two notes:

1. I added a small section on Updating the Workflow Browser, since I foresee that we will likely want to involve UI towards the end of this build for some small updates to the browser.
2. Let's get this merged! There's a whole bunch of open comments on this PR, but I don't think its necessary to resolve all of them before merging and proceeding to Build. If there's any glaring topics you would like resolved before merge, please let me know.

Panaetius (approved):

mohammad-sdsc (approved):

feature-pitches/002-file-based-workflows/002-file-based-workflows.md
lokijuhy: Yes

PR #17, opened 2022-09-20
chore: move to rfc folder

olevski:
This was an old RFC that ended up being placed in the wrong place. Now I moved it in the folder with the rest of the RFCs.


rokroskar:
oops I forgot to check that in my review 🤦

rokroskar (approved):

PR #20, opened 2022-12-11
feat: gateway refactor RFC

olevski:
This has two parts:
1. Refactoring and streamlining the gateway
2. Implementing a centralized authentication service for sessions


rfcs/010-gateway-refactor/010-gateway-refactor.md
```diff
@@ -0,0 +1,205 @@
...
+
+### Centralized Jupyter server authentication
+
+### Diagrams
+
+1. Creating a user server
+
+```mermaid
+graph TD
+    Ingress -- 1<br>Session ID --> Gateway
```
rokroskar, line 66: I guess `ingress` is the user / client?

rfcs/010-gateway-refactor/010-gateway-refactor.md
```diff
@@ -0,0 +1,205 @@
...
+    subgraph GatewayGroup[ ]
+    Gateway
+    DB[(DB)]
+    Gateway -- 2<br>Session ID --> DB
+    Gateway -- 3<br>Server ID --> DB
+    DB -- 4<br>KC access token<br>Server ID --> Gateway
+    end    
+    
+    Gateway -- 5<br>KC access token<br>Server ID--> Notebook[Notebook service]
+    Notebook -- 6<br>KC user ID<br>Server ID--> Server[User server]
```
rokroskar, line 77: What are KC tokens used for if the access to servers is brokered by the gateway via session tokens?

rfcs/010-gateway-refactor/010-gateway-refactor.md
```diff
@@ -0,0 +1,205 @@
...
+    Ingress -- 1<br>Session ID --> Gateway
+    
+    subgraph GatewayGroup[ ]
+    Gateway
+    DB[(DB)]
+    Gateway -- 2<br>Session ID --> DB
+    DB -- 3<br>KC access token --> Gateway
+    end    
+    
+    Gateway -- 4<br>KC access token --> Proxy
```
rokroskar, line 93: so there still is a proxy inside the session pod? That processes the JWT? The intro says there's no more oauth2 proxy in the sessions pod so I'm confused.

rfcs/010-gateway-refactor/010-gateway-refactor.md
```diff
@@ -0,0 +1,205 @@
...
+
+    subgraph Server[User server]
+    Proxy -- 5 --> Jupyter
+    end
+```
+
+3. User server sidedcar accessing a protected service
```
rokroskar, line 100:
```suggestion
3. User server sidecar accessing a protected service
```

rfcs/010-gateway-refactor/010-gateway-refactor.md
```diff
@@ -0,0 +1,205 @@
...
+    DB -- 4<br>KC access token<br>Git access token --> Gateway
+    Gateway -- 5<br>KC access token<br>Git access token<br>Sticky session --> Core
+```
+
+5. Requests from the ui-server
+
+Here the gateway will simply require fully valid Keycloak access tokens
+(which the ui-server already has). If this is implemented, the ui-server
+will be the only component that will be coming to the gateway with Keycloak
+access tokens. All others will be coming with a session ID. So to further
```
rokroskar, line 146: seems like the ui-server should conform to this architecture instead?

rfcs/010-gateway-refactor/010-gateway-refactor.md
```diff
@@ -0,0 +1,205 @@
...
+## Unresolved questions
+
+Issues to be resolved through the RFC process before this is merged:
+- agreement on the resposiblities of the gateway
+- agreement on the design and decisions
+- whether to use more persistent storage for some sessions that need to be
+longer-lived (i.e. sessions for Jupyter servers)
+
+Issues to resolve through the implementation of this RFC:
+- potential elimination of traefik in the gateway (all golang web frameworks support reverse proxying out of the box)
```
rokroskar, line 198: oh i was reading this thinking that it assumes Traefik is gone. Which adds significantly to the complexity of the implementation because now we have to also re-implement all the logic that is inside the middleware definition - not a bad thing because it's notoriously difficult to understand what is going on in there. otoh is implementing yet another proxy really something we want to be doing?

rokroskar (commented):
Nice proposal @olevski, thanks!

A couple of questions (in addition to the ones in-line): 

- where/when does the gateway get other tokens like the gitlab token? Should we envision a future here where those are handled by a dedicated secret-handling service? 
- it looks like there is still a slightly confusing situation in which some services handle some secrets themselves (e.g. core service making requests to gitlab) and in other cases the gateway is injecting the credentials into the request (e.g. sidecar accessing gitlab). Imho it would be nice to have one approach, keeping in mind that being more conservative about sending credentials around is better. 

For me the main thing that needs to be _very_ clearly resolved in this RFC is what pattern we expect for credentials handling. Should the gateway give the credentials to the services as needed or should it inject them into the requests? I think we should commit to one or the other but not both. Once this is done, _all_ services should use this pattern, including the KG services that are currently handling tokens for themselves.

rfcs/010-gateway-refactor/010-gateway-refactor.md
olevski: @rokroskar I need to correct the intro. We need something either in the gateway or in the jupyter server to decide whether the request coming into to the jupyter server is from the right user. IMO this is an authorization question - not authentication, as such it makes sense to not make this the responsibility of the gateway. In this situation the gateway simply authenticates the user and injects the user's access token in the request. Then the jupyter server proxy can decide if that user should be granted access to the session or not.

If this is moved in the gateway the gateway would need to somehow figure out which user is the owner of the jupyter server being accessed. This means that we need a database with this info or the gateway would need to query k8s. I think that for this reason and as I mentioned above this should not be the responsibility of the gateway. When we implement a central authorization service this can be offloaded there. But for now it makes the most sense for this to remain in the jupyter server. I think this can be fulfilled by a container in the jupyter server that either has traefik with a simple config or some kind of configurable proxy. It is possible that the oauth2 proxy can be configured to handle this. But we can see and decide what is easiest and makes most sense.

rokroskar: ok that's fine and that's the way I understood it from the diagrams, but please make it consistent (I realize this is evolving so no worries :)

rfcs/010-gateway-refactor/010-gateway-refactor.md
olevski: Yes exactly. 

I had `browser --> ingress --> gateway` everywhere but I removed the browser part to save on space.

rfcs/010-gateway-refactor/010-gateway-refactor.md
olevski: There are a few reasons for this:
1. The Keycloak access tokens can be independently validated by every service. Also every service can have a set of claims and/or audiences it accepts. This makes things safer where even if a request from within the cluster went to a service (bypassing the gateway) and that request is not authenticated the services will deny it.
2. In this particular case the notebook service needs user information such as user ID, name and email that can be extracted from a Keycloak ID token. I will correct this to say ID token.

rfcs/010-gateway-refactor/010-gateway-refactor.md
olevski: Yes but the ui-server already has sessions with each user and it keeps access tokens for these sessions up to date. Removing this functionality from the ui-server and merging it with the gateway was too big of a change IMO. And it is something that I am not sure whether we should or should not do. So accepting KC access tokens in the gateway  was the simplest thing to do. 

The alternative to this is getting the ui-server to have a session with the gateway. But there would need to be one such session per user. Not just one for the whole service. I think this definitely complicates things and makes it so that we need changes in the ui-server in addition to the gateway.

rfcs/010-gateway-refactor/010-gateway-refactor.md
olevski: IMO we use traefik only for very simple path rewriting and cookie stripping. So doing this in the golang proxy is really not hard. The golang server framework I would like to use (https://echo.labstack.com/) has very easy to use middlewares that give you much more flexibility and expressiveness than the traefik configuration. In addition we can also write unit tests for these whereas we cannot unit test the traefik configuration.

Lastly when we want sticky sessions we have to selectively remove some cookies in the gateway (but leave the sticky session cookie in there). Currently we remove all of them. To keep some cookies in a request but remove all others we need to use a custom plugin for traefik. I wrote one for this purpose [here](https://github.com/SwissDataScienceCenter/cookiefilter). I imagine that we may find ourselves in similar situations down the road as well where the functionality we need is not provided by traefik and we will have to write our own plugins. If we use the golang server framework to do this proxying then we do not have to worry about this.

olevski:
> where/when does the gateway get other tokens like the gitlab token? Should we envision a future here where those are handled by a dedicated secret-handling service?

Currently the gateway gets the gitlab access token when the user logs in. The redirect url for Gitlab points at the gateway. And when the gateway receives a request here it finishes the oauth flow and gets a Gitlab refresh token and a gitlab access token. So yes this can be separated out in the future. But perhaps a secret service would handle cases where users enter other credentials in Renku that do not expire (unlike oauth access or refresh tokens).  I think that handling the login from of a user and collecting all the tokens from said oauth login flow should be responsibility of the gateway. 

When we add a secret service I envision that only the gateway would be allowed to talk to it potentially. But we can figure this out in a separate RFC. Any Oauth tokens which require interaction from the user to acquire and which require periodic renewal should be managed by the gateway IMO.

olevski:
> it looks like there is still a slightly confusing situation in which some services handle some secrets themselves (e.g. core service making requests to gitlab) and in other cases the gateway is injecting the credentials into the request (e.g. sidecar accessing gitlab). Imho it would be nice to have one approach, keeping in mind that being more conservative about sending credentials around is better.

I absolutely agree. I will check how the core service gets their gitlab tokens. I thought it already went through the gateway.

I thought that only the KG managed copies of the gitlab tokens. And that can also be moved to the gateway. For example when a project is created the kg can have a long lived session with the gateway (similar to the jupyter servers). Or alternatively the gateway can be modified to inject an appropriate access token for the correct Gitlab user/project for requests coming from the KG. I did not go into too much details on that here because I wanted to keep the scope of the RFC smaller.

rfcs/010-gateway-refactor/010-gateway-refactor.md
olevski: Done. The intro states this more clearly now.

rfcs/010-gateway-refactor/010-gateway-refactor.md
lorenzo-cavazzi: I agree that the UI server should conform: it should be treated as a client like the others (_there might be a few complications so I'm happy if we postpone it to phase 2 and initially accept valid access tokens_).

We can handle it by keeping the session with the client, and the server will map that to another session with the gateway. This way, we will just attach/swap the session cookies to every request instead of swapping with the user token.
Ideally, the change is not that big on our side.

rfcs/010-gateway-refactor/010-gateway-refactor.md
```diff
@@ -0,0 +1,211 @@
...
+- User servers (i.e. the sidecar container needs be able to swap some kind of ID
+or access token for a Gitlab token)
+- Actual users logging in through the UI
+- Actual users logging in through the CLI
+
+Using JWT access tokens for this purpose is not feasible. This is because access
+tokens from Keycloak expire usually within an hour or several hours. This means
+that anyone who has to remain authenticated for a long time with the gateway
+needs to also get a refresh token and worry about using this to have valid access
+tokens. Accepting an expired access token is not secure.
```
wesjdj, line 43: Maybe I've misunderstood, but what would be the problem with using the refresh token to get a new access token regularly?

rfcs/010-gateway-refactor/010-gateway-refactor.md
```diff
@@ -0,0 +1,211 @@
...
+
+The gateway should instead have server-side sessions with every "client". For each such
+session the gateway will issue a random, hard to guess ID that will be stored either
+in a cookie or simply as a secret. This is exactly how sessions are handled by the
+ui-server. The cookie or ID is presented with every request to the gateway. The gateway
+looks this ID up in redis and if there is a match then any tokens/credentials stored
```
wesjdj, line 49: Would it make sense to use something like HashiCorp Vault instead of Redis?

rfcs/010-gateway-refactor/010-gateway-refactor.md
olevski: There can be a few issues with this:
- Some identity providers (keycloak by default does not do this) will make a refresh token invalid as soon as you have used it once. So by distributing refresh tokens to all services we get into a situation where we can have a race condition and have only a single service successfully refresh while all others are screwed.
- Distributing the refresh tokens in addition to the access tokens in many different places is less secure by storing them only in a single place and having only a single service access them.

rfcs/010-gateway-refactor/010-gateway-refactor.md
olevski: We may end up using Vault for storing static secrets from users such as S3 keys, or database credentials, etc. But I think using Vault for this is overkill. Storing secrets in redis in the backend for a user session is standard IMO. Also not sure if Vault will be as performant as Redis when it comes to having many requests that continuously and periodically check all tokens and refresh them if needed. This is probably not that big of a deal for secrets that are static, but it may cause problems for oauth access/refresh tokens.

olevski: @wesjdj have you seen or used of Vault to store oauth2 tokens? I will do some googling to see if this is feasible.

wesjdj: @aledegano might have some more insight on this but I think Vault is quite performant and can store whatever is asked of it

rfcs/010-gateway-refactor/010-gateway-refactor.md
```diff
@@ -0,0 +1,211 @@
...
+
+## Summary
+
+This proposes a full rewrite of the current gateway service. In addition, it also proposes
+changes to the way that access control to Jupyter servers is handled. Currently, each Jupyter
+server establishes a session with the user. But going forward each user will have a session
+with the gateway which will be used to control access into each Jupyter server. 
+The main motivation for changing the access control for Jupyter servers is the desire to run
+each Jupyter server in its own subdomain. The problem with unique subdomains is that each Jupyter
+server's subdomain would have to be registered in Keycloak so that the Oauth2-proxy in each
```
seanrmurphy, line 14: What do we mean here by subdomain? Do we think we will have an fqdn of `ssh.my-session.renkulab.io` or just `my-session.renkulab.io`?

rfcs/010-gateway-refactor/010-gateway-refactor.md
```diff
@@ -0,0 +1,211 @@
...
+
+Reasons for rewriting the gateway:
+- the current code base is hard to maintain
+- one service could replace both traefik and the gateway
+- problems with managing sessions across different services/clients
+- using golang should result in more efficient use of resources and improved performance
+
+Reasons for centralized session authentication:
+- running sessions in subdomains which would allow for SSH access
+
```
seanrmurphy, line 28: Could we add something on Requirements? Does the following seem reasonable?

## Requirements

We currently have two separate issues in which the gateway plays an integral part:
- managing access from external entities (browsers, command line tools) to Renku resources
  - we need to ensure that all requests hitting our internal endpoints have appropriate identities and authorizations
- supporting access from (internal) Renku services to external protected entities
  - in this case we are using tokens generated by external services and managed by Renku to access these resources

In general, we should limit token distribution such that Renku services have them only as necessary, especially for refresh tokens. We should also be aware of the privileges associated with tokens we manage, specifically for tokens relating to external services.

We should understand what happens if we lose all tokens and should be able to recover from this situation. We should also have some basic understanding of the implications of being compromised and what could happen if all tokens which we manage become untrusted.

rfcs/010-gateway-refactor/010-gateway-refactor.md
olevski: It would probably be something like `my-session.sessions.renkulab.io`. I don't think we need to put `ssh` in the host name. SSH would just by default be on port 22 and the notebook would be on 80 or 443. So we do not need the `ssh` part at all. But we do want to have the `sessions` part preceding the name of the sessions just to separate things out properly.

rfcs/010-gateway-refactor/010-gateway-refactor.md
olevski: > if we lose all the tokens

@seanrmurphy I am not sure what do you mea by this. Are you referring to the tokens that we expect to be persisted suddenly not being available?

> of the implications of being compromised and what could happen if all tokens which we manage become untrusted.

Can you give an example scenario here? Because this seems really broad in order to address. 

Also there is a difference between the session IDs we will be issuing and the Keycloak or Gitlab access or refresh tokens. By default we do not trust any sessionID - unless we can find it in the database. The keycloak access and refresh tokens can be verified with the keycloak public signing key. The keycloak and the gitlab tokens can also all be introspected with keycloak and gitlab respectively. But if someone gains access to the database where the tokens are stored then there is not much that we can do other than shut everything down and let users know. We will encrypt the stuff in the database. But I think we will have to have the encryption key in the gateway service so that we can actually use the tokens. So it possible that an attack could get access to the encryption key since we will using it all the time anyhow.

rfcs/010-gateway-refactor/010-gateway-refactor.md
olevski: So on this topic I stumbled upon this issue on vault: https://github.com/hashicorp/vault/issues/96 which then lead me to this vault plugin: https://github.com/puppetlabs/vault-plugin-secrets-oauthapp. This makes things more interesting because with the plugin vault would automatically refresh the oauth tokens we store. But without the plugin we would have to do this ourselves.

But I know very little about vault and have never used in production. The perspective on this from someone with more experience would be definitely appreciated.

rokroskar: oh that's a very interesting option! Some vault ninja should try it against the dev gitlab...

wesjdj:
One possibility I think is worth exploring is not using the whole session token thing for GitLab anymore. I would be interested to know if any RenkuLab deployments couldn't use a GitLab admin personal access token. And, if there are, we have people create a Renku user and invite them to whichever projects they would like to use Renku on. I wouldn't say I like the session token stuff at all, to be honest.

rfcs/010-gateway-refactor/010-gateway-refactor.md
olevski: > How about having a single service responsible for refreshing the token then and storing the access token somewhere centralised like Vault instead?

That is exactly the idea.

olevski:
> One possibility I think is worth exploring is not using the whole session token thing for GitLab anymore. I would be interested to know if any RenkuLab deployments couldn't use a GitLab admin personal access token. And, if there are, we have people create a Renku user and invite them to whichever projects they would like to use Renku on. I wouldn't say I like the session token stuff at all, to be honest.

Can you give a bit more details on how the admin's personal access token will be used? I am very scared of giving out the admin's token to any service or to a jupyter server. This means that if an attacker gains access to this they gain full access to absolutely all data we store. Also not sure if the admin can see gitlab users' secrets they have stored in the CI.

On the same topic, what happens when we decouple from our hosted Gitlab and start to support Gitlab.com and Github.com? And how does then this work for someone else deploying Renku where they want Renku to work with Gitlab.com and/or Github.com?

wesjdj:
> > One possibility I think is worth exploring is not using the whole session token thing for GitLab anymore. I would be interested to know if any RenkuLab deployments couldn't use a GitLab admin personal access token. And, if there are, we have people create a Renku user and invite them to whichever projects they would like to use Renku on. I wouldn't say I like the session token stuff at all, to be honest.
> 
> @wesjdj can you give a bit more details on how the admin's personal access token will be used? I am very scared of giving out the admin's token to any service or to a jupyter server. This means that if an attacker gains access to this they gain full access to absolutely all data we store. Also not sure if the admin can see gitlab users' secrets they have stored in the CI.
> 
> On the same topic, what happens when we decouple from our hosted Gitlab and start to support Gitlab.com and Github.com? And how does then this work for someone else deploying Renku where they want Renku to work with Gitlab.com and/or Github.com?

I can't, really, because I find it very hard to follow all the different connections between Renku components and GitLab. Maybe what I'm suggesting is more relevant to KG. Either way, I think we need more of an architectural review on how we interact with GitLab or whatever backend we chose to use. The session token thing just does not feel like a good way of solving whatever it is we're trying to do to me.

rfcs/010-gateway-refactor/010-gateway-refactor.md
aledegano: I can try that when I'm back from x-mas holidays.

rfcs/010-gateway-refactor/010-gateway-refactor.md
seanrmurphy: Related to the ssh issue; have we ever discussed something like the gitpod ssh gateway which would mean hosting only a single ssh entry point for all users?

https://www.gitpod.io/docs/configure/user-settings/ssh#ssh-gateway

rfcs/010-gateway-refactor/010-gateway-refactor.md
seanrmurphy: > @seanrmurphy I am not sure what do you mean by this. Are you referring to the tokens that we expect to be persisted suddenly not being available?

It relates to redis and requirements: we can impose a requirement that tokens be persisted (for how long?) or we can think of disposable tokens. From our discussion earlier, I think that you basically assume tokens will be persisted.

That said, if someone on the operations team screwed up and lost all the tokens, would there be BAD implications? I guess in reality, it would prob mean some friction/inconvenience for users but it would not render the platform unusable.

> Can you give an example scenario here? Because this seems really broad in order to address.

Being compromised is of course a risk and there are many variants of it. To be systematic, we could break it down into different scenarios (which could be worth doing - eg one refresh token is snooped, one access token is snooped, all tokens in one service are snooped etc), but for now, let's take the worst case: an attacker gets access to all the tokens which we are managing - what would be the implications here? I guess she would be able to impersonate all Renku users (on all projects that user has access to?)...would she be able to delete all projects the users have access to (and demand some ransom)? For me, the answer here does not need to be definitive, but we should be aware of and document at least the envelope of the risk profile we consider...

rfcs/010-gateway-refactor/010-gateway-refactor.md
```diff
@@ -0,0 +1,211 @@
...
+The gateway should instead have server-side sessions with every "client". For each such
+session the gateway will issue a random, hard to guess ID that will be stored either
+in a cookie or simply as a secret. This is exactly how sessions are handled by the
+ui-server. The cookie or ID is presented with every request to the gateway. The gateway
+looks this ID up in redis and if there is a match then any tokens/credentials stored
+under that key can be added to requests initiated by that ID.
+
+These sessions are managed/started as follows:
+- when a user logs in they get a session
+- when a user server is started the server gets a redis-based session
```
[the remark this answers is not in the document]
olevski: There is no difference.

olevski:
> I can try that when I'm back from x-mas holidays.

@aledegano I am not exactly sure what you are referring to here?

aledegano:
> > I can try that when I'm back from x-mas holidays.
> 
> @aledegano I am not exactly sure what you are referring to here?

I was replying to Rok's comment about creating a Demo of Vault+the OAuth plugin.
You can see the correct chain of comments in the review page, but here it got lost...

rfcs/010-gateway-refactor/010-gateway-refactor.md
olevski: Ok I had to google around a bit to find the repo for that and then find the relevant code. But it is here. So that is a possibility.

https://github.com/gitpod-io/gitpod/tree/main/components/ws-proxy/pkg/sshproxy

rfcs/010-gateway-refactor/010-gateway-refactor.md
olevski: So if the tokens are lost and the persisted tokens in the database is lost that would mean that:
- kg integration would fail but for new users who log in it should start to work
- re-provisioning of the kg triples would never complete 100% but for whoever has logged in since it will work
- users would have to log in again in the UI
- the notebook service, the sidecar and core service would also not be able to function fully for existing servers (or cached projects for the core service) but this should in theory be able to work again when users logged back in with the ui

If the attacker gets access to all tokens we manage they would be able to impersonate all users and not just read users' data from a repo but also push. So they could do a series of force pushes to delete data. Perhaps branch protection rules may help here but this is not something we want to be relying on. So the consequences are really bad.

rfcs/010-gateway-refactor/010-gateway-refactor.md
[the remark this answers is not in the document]
olevski: no difference

rfcs/010-gateway-refactor/010-gateway-refactor.md
```diff
@@ -0,0 +1,211 @@
...
+
+The gateway has to have sessions with different clients. Some of these are:
+- User servers (i.e. the sidecar container needs be able to swap some kind of ID
+or access token for a Gitlab token)
+- Actual users logging in through the UI
+- Actual users logging in through the CLI
+
+Using JWT access tokens for this purpose is not feasible. This is because access
+tokens from Keycloak expire usually within an hour or several hours. This means
+that anyone who has to remain authenticated for a long time with the gateway
```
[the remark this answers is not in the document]
olevski: > 2. whomever it is in the above point, can we just automate the renewal when the tokens are about to expire?

Yes that is very much the plan.

> 3. are these tokens used by users independently from the context of Renku authenticating between services?

Users currently do not handle the keycloak or Gitlab access tokens directly. The closest they get to this is when they have a session with the oauth2 proxy in every jupyter server. The oauth2 proxy uses client-side encrypted cookie based sessions. And that session contains either the full keycloak access token or some portion of it but encrypted. But still the users (and here by "users" I mean people clicking around in the ui on renkulab) do not refresh, acquire or in any way handle any access tokens.

rfcs/010-gateway-refactor/010-gateway-refactor.md
```diff
@@ -0,0 +1,211 @@
...
+needs to also get a refresh token and worry about using this to have valid access
+tokens. Accepting an expired access token is not secure.
+
+The gateway should instead have server-side sessions with every "client". For each such
+session the gateway will issue a random, hard to guess ID that will be stored either
```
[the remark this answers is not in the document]
olevski: Ok yeah I added the link in the text where I mention the ID.

rfcs/010-gateway-refactor/010-gateway-refactor.md
```diff
@@ -0,0 +1,211 @@
...
+under that key can be added to requests initiated by that ID.
+
+These sessions are managed/started as follows:
+- when a user logs in they get a session
+- when a user server is started the server gets a redis-based session
+- when a login happens through the CLI there is a also a session
+- anonymous users also have redis-based sessions but these are empty
+since there is no additional credentials that need to be managed for anonymous users
+
+The gateway then has the following responsibilities:
```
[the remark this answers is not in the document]
olevski: Yes. I replaced the bullet points

rokroskar:
> And, if there are, we have people create a Renku user and invite them to whichever projects they would like to use Renku on.

@wesjdj I've done some more thinking about this and I believe this could be a reasonable approach to solve some of our problems (we still need a user's token to push to their repo, for example, but not to fetch the commits and process them for the purposes of the KG). I think we can automate adding the user to a project to reduce the friction as much as possible. For the most part, this seems to be the expected way for some variation of "apps" on GitLab - seems to be more streamlined/hidden on GitHub. As long as users don't complain about this random user being a member of their project, it should be fine.

rfcs/010-gateway-refactor/010-gateway-refactor.md
rokroskar: The ssh gateway seems to make sense. Though is the main purpose to simplify ingress and gain visibility on active ssh connections? What I'm wondering is if it's solving something we're not realizing should be a top concern, or if it's a nice-to-have.

seanrmurphy: I think it would be a 'less moving parts' solution, which I generally prefer - I also like the visibility dimension.

If it's something that works out of the box, then I think it would save time/effort; if it's something that requires extensive customization, then I think I guess more consideration is required.

rfcs/010-gateway-refactor/010-gateway-refactor.md
Panaetius: > or cached projects for the core service
I don't think this is true. Since we get the token in each request, and a request means a user has logged in again, it shouldn't affect anything. It'd only break e.g. a long running job like dataset import that was started before the issue and tries to push after it.

rfcs/010-gateway-refactor/010-gateway-refactor.md
seanrmurphy: As discussed with @olevski, I spent a bit of time looking at this last Fri - I surveyed a few projects but boiled it down to the gitpod proxy and sshr. I spent time looking at `sshr`. It more or less works and does what it indicates; however, this solution has a fundamental limitation that it is essentially gluing two ssh sessions together and there needs to be some thought around how this can work and what the implications are. (There were issues that many projects on github in this vein assume use of ssh proxying but that is obviously some friction we don't want to push to users right now).

I looked a bit at the gitpod proxy; I built it but did not really try to get it working - for me, it was quite coupled to the gitpod ecosystem and pulled down keys from a central service, iiuc. @olevski has managed to pull out the base functionalty; however.

I set up `sshr` and managed to get a variant of this working - the project is a few years old, so I had to update some libs. However, I realized that in the demo, there were private keys both on the client end and within the proxy.

I then started to think about this a little further and I guess we would end up with the same solution as gitpod - basically, having pubkeys on the proxy and some authentication keys managed by us which are used between the proxy and the session. The purpose of these is to ensure that users cannot log into each other's sessions.

That's as far as I got on this last week; happy to discuss further if we wanna pursue this.

wesjdj:
> > And, if there are, we have people create a Renku user and invite them to whichever projects they would like to use Renku on.
> 
> @wesjdj I've done some more thinking about this and I believe this could be a reasonable approach to solve some of our problems (we still need a user's token to push to their repo, for example, but not to fetch the commits and process them for the purposes of the KG). I think we can automate adding the user to a project to reduce the friction as much as possible. For the most part, this seems to be the expected way for some variation of "apps" on GitLab - seems to be more streamlined/hidden on GitHub. As long as users don't complain about this random user being a member of their project, it should be fine.

As for our new approach for KG, that is pretty much exactly what we do, as the automatically generated Renku user for the project access token is visible in the project

rfcs/010-gateway-refactor/010-gateway-refactor.md
seanrmurphy: One other point which was not completely clear to me - can we assume that the username on the session is constant? In the gitpod case, the username is always `gitpod`. I guess we have to assume this...

seanrmurphy: Probably the simplest solution for now would be to have no passwords/keys in the user session and use network policies to block access to ssh on user sessions from everything except the proxy. The user's pubkey gets pushed to the proxy prob via a configmap which links session name to pubkey and if you log in to the proxy successfully, you get forwarded to the remote session.

This is not an ideal solution, but prob the quickest to implement and prob sufficiently secure for our needs.

seanrmurphy: @Panaetius - this is specifically focused on the proxying solution (if we choose to go that way).

If the user's pubkeys are inside the session, then the proxy would need the user's private key which is not really a runner. Iiuc, in gitpod, they create another set of pub/priv keys just for access from the proxy to the session but these only live inside the gitpod context (ie they are generated by gitpod and looked up by the proxy on session initialization - the user can prob see the pubkey as it lives in the session but never sees the corresponding privkey).

If we don't use the proxying solution, then I think the natural solution is as you suggest.

rokroskar: @seanrmurphy where is the user's pub key? Presumably the proxy has it but how does it get it?

rokroskar: nvm, I see you said above via a configmap - but this needs to be made dynamic somehow?

seanrmurphy: > nvm, I see you said above via a configmap - but this needs to be made dynamic somehow?

I guess a configmap is most natural; of course, it could be possible to write another service which handles this. 

If we went with a configmap approach, it would need to be dynamic.

Panaetius: ok, thanks for the clarification.

> One other point which was not completely clear to me - can we assume that the username on the session is constant? In the gitpod case, the username is always gitpod. I guess we have to assume this...

Right now it's usually `jovyan` and I think there's some parts of notebooks service that depend on this, but it's not really an enforced constraint. 

Also the session URLs depend on username and commit (since you can start multiple sessions for the same user), and we discussed having an additional FQDN for SSHing that can only point to one of those sessions at a time. Users could still use the FQDN of a specific session if they want, but the persistent FQDN makes their lives easier since they wouldn't have to update the connection information every time they start a new session.

I think having an ssh proxy should be separate from the gateway refactor and is something we can add in the future. For now we'd just forward whatever traffic and it's up to the sshd in the session to do auth. I have a test project with sshd set up this way [here](https://dev.renku.ch/projects/ralf.grubenmann/ssh-into-sessions/files). Though that means that the gateway doesn't do auth for port 22 traffic. But I think this would be the fastest thing to get working, and we can add a sshgateway later on as a separate feature.

olevski: I am resolving this because the ssh proxy is out of scope for this RFC. There will be a separate RFC for that.

rfcs/010-gateway-refactor/010-gateway-refactor.md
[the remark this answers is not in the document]
olevski: > who is "anyone that has to remain authenticated", services, users, both? Are you referring to the "different clients" from the above text block?

By anyone I mean either users or internal services that need to act on behalf of a user.

Panaetius:
We should get this merged, since it's essentially done already

Panaetius (approved):

PR #21, opened 2022-12-16
feat: dedicated renku git user

olevski:
As @wesjdj suggested.

This idea is definitely worth exploring in a bit more detail. And we have been dismissing it perhaps a bit too early before.

### DIFF
--- a/rfcs/011-dedicated-renku-git-user/011-dedicated-renku-git-user.md
+++ b/rfcs/011-dedicated-renku-git-user/011-dedicated-renku-git-user.md
@@ -1,0 +1,113 @@
+- Start Date: 16-12-2022
+- Status: Proposed
+
+# Dedicated Gitlab User for Knowledge Graph
+
+## Summary
+
+Every Renku deployment should have a dedicated regular user in every Gitlab it is connected
+with. This user will be called something like `Renku` or `Renkubot` and it will be used
+by the KG to access the project and process it's history.
+
+## Motivation
+
+The KG needs authenticated access to Gitlab so that it can analyze and index a project's 
+Git history. Currently the KG has to store many Gitlab user tokens in postgres. 
+These tokens are used depending on what project the KG is processing events for. This means
+that the KG has to maintain and manage a database with Gitlab acces tokens (or project tokens).
+If these are lost then the KG would not be able to access this project to add it in its database
+during reprovisioning. 
+
+In addition with the switch from Gitlab user access tokens to project tokens there was a
+lot of automatically generated "bot" users that Gitlab created when the project tokens
+were created. This is just what happens when a Gitlab project token is created but it may
+be cause concerns with some admins or users.
+
+Lastly, with having a regular user being dedicated to the KG, there is no more impersonation
+of users and "fake" user activity that comes from the periodic checks and indexing of projects
+that is done by the KG. All this activity will now be associated with the Renkubot user rather
+than individual real users.
+
+## Design Detail
+
+### Helm Chart
+
+In the Helm chart we would add a hook that would create a user in the Gitlab deployment
+that Renku is connected to. The username and password will be saved in a secret. The KG
+would then have access to the secret and use it to authenticate with Gitlab.
+
+Alternatively the Helm chart can also accept existing user credentials. 
+
+### Gateway
+
+Either as part of this RFC or as an improvement the management of this secret / tokens for 
+the Renkubot user should be done by the Gateway and the KG will simply either ask the Gateway 
+for credentials or a Gateway's forward proxy will inject the right credentials in the right calls.
+
+This will also be helpful when we support multiple Gitlabs/Github. In this case the KG does
+not need to keep track which credentials to use for which Gitlab. The gateway can figure this
+out and inject the right credentials for the Renkubot user.
+
+### UI
+
+There are some changes needed in the UI for this. Namely:
+- To get a list of all projects for a user the UI would have to call Gitlab on behalf
+of the user. This is because the KG will now only know about projects to which the Renkubot
+user has been added. In other words, the KG will not know about any projects for which 
+KG integration is not active. It could figure this out for public Gitlab projects
+but it would be impossible to find private projects that are not in the KG.
+- Creating a project on Renku will have to include adding this Renkubot user to the project
+with `Reporter` permissions.
+- KG integration for a specific project can still be verified with the KG. In some cases
+the KG will simply not know anything at all for a project. In this case KG integration
+for this project is not enabled. Checking that the Renkubot user is in the project with the
+right permissions is not enough. It could be that the project has the Renkubot user but
+the KG has not fully finished indexing the project.
+- Enabling KG integration on projects that do not have it will have to be modified to also
+add the user to the project similarly to creating a project.
+- Disabling / removing KG integration results in removal of the project from the KG and removing
+the Renkubot user from the project.
+
+### KG
+
+Depending on when the Gateway refactoring is completed, the KG may not have to deal with the Renkubot
+credentials at all. This is the preferred option. But it may turn out that an initial implementation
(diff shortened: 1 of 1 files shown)


rfcs/011-dedicated-renku-git-user/011-dedicated-renku-git-user.md
```diff
@@ -0,0 +1,113 @@
...
+for credentials or a Gateway's forward proxy will inject the right credentials in the right calls.
+
+This will also be helpful when we support multiple Gitlabs/Github. In this case the KG does
+not need to keep track which credentials to use for which Gitlab. The gateway can figure this
+out and inject the right credentials for the Renkubot user.
+
+### UI
+
+There are some changes needed in the UI for this. Namely:
+- To get a list of all projects for a user the UI would have to call Gitlab on behalf
```
jachro, line 54: Actually, I think this might work without any changes as long as we get the user OAuth access token passed to our User's Projects API as we do now. The reason is that we find the list of user's non-active projects by calling the GL User Projects API with both the userId and OAuth token we get from the UI in the request (these two are injected by the Gateway). I mean, at least now this API doesn't use the RenkuBot user (or Project Access Token) and probably it would be hard to achieve the same functionality if it would have to.

rfcs/011-dedicated-renku-git-user/011-dedicated-renku-git-user.md
```diff
@@ -0,0 +1,113 @@
...
+of the user. This is because the KG will now only know about projects to which the Renkubot
+user has been added. In other words, the KG will not know about any projects for which 
+KG integration is not active. It could figure this out for public Gitlab projects
+but it would be impossible to find private projects that are not in the KG.
+- Creating a project on Renku will have to include adding this Renkubot user to the project
+with `Reporter` permissions.
```
jachro, line 60: It might also be that the Renkubot user would be added to the project during the activation call on the KG side. Obviously, it's not a must, but I can see that building a coherent picture and keeping the responsibilities in the proper places. I might be wrong but for me, it's a part of the activation process.

rfcs/011-dedicated-renku-git-user/011-dedicated-renku-git-user.md
```diff
@@ -0,0 +1,113 @@
...
+- KG integration for a specific project can still be verified with the KG. In some cases
+the KG will simply not know anything at all for a project. In this case KG integration
```
jachro, line 62: Yes, that I see matching nicely to the check we do on `GET /api/projects/24541/graph/status`. At the moment it returns info about activation status and processing progress.

rfcs/011-dedicated-renku-git-user/011-dedicated-renku-git-user.md
```diff
@@ -0,0 +1,113 @@
...
+for this project is not enabled. Checking that the Renkubot user is in the project with the
+right permissions is not enough. It could be that the project has the Renkubot user but
+the KG has not fully finished indexing the project.
+- Enabling KG integration on projects that do not have it will have to be modified to also
+add the user to the project similarly to creating a project.
+- Disabling / removing KG integration results in removal of the project from the KG and removing
+the Renkubot user from the project.
```
jachro, line 69: That might be done when we remove a project from KG. At the moment we try to remove the webhook but we could easily try to remove the Renkubot user, too. The only question is whether we can do that using the Renkubot token? It's also worth mentioning, that currently, there's no route of disabling/removing a project from KG.

rfcs/011-dedicated-renku-git-user/011-dedicated-renku-git-user.md
```diff
@@ -0,0 +1,113 @@
...
+
+### KG
+
+Depending on when the Gateway refactoring is completed, the KG may not have to deal with the Renkubot
```
jachro, line 73: I guess it's not a big difference for the KG which option we'd choose. At least from the implementation side, I think it might be even easier if we'd get the credentials injected so we can simply exchange them for an OAuth token in the token repository.

rfcs/011-dedicated-renku-git-user/011-dedicated-renku-git-user.md
```diff
@@ -0,0 +1,113 @@
...
+credentials at all. This is the preferred option. But it may turn out that an initial implementation
+of this gives the Renkubot credentials directly to the KG at first.
+
+## Drawbacks
+
+1. The Renkubot user will have read access to many repos. If the credentials are stolen or leaked
```
jachro, line 79: Actually, I could understand if this would be a serious concern for some of our users or admins.

rfcs/011-dedicated-renku-git-user/011-dedicated-renku-git-user.md
```diff
@@ -0,0 +1,113 @@
...
+then they could be used to read the contents of private repos. 
+
+2. When this is rolled out we would automatically add the Renkubot user to all projects. This may
+be confusing to users.
```
jachro, line 83: Well, all the repos already have a renkubot user :) However, each repo has a different one.

rfcs/011-dedicated-renku-git-user/011-dedicated-renku-git-user.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Unresolved Questions
+
+1. If a user is added to a project in Gitlab do they have to click a link in their email
+to accept the invitation? If this is the case then this whole proposal is not feasible.
+
+2. Do we implement simultaneously with the gateway refactoring or do we have an initial
+implementation where the KG handles the Renkubot credentials.
+
+3. Should we add the Renkubot users to existing projects when this is deployed for the
```
jachro, line 112: We could do that if using the Project Access Tokens we already have allow doing so.

Panaetius:
should change the folder, resource access control is already 011

rfcs/011-dedicated-renku-git-user/011-dedicated-renku-git-user.md
rokroskar: All project maintainers will probably be notified that the bot user has been added. We will need to communicate this clearly to the users so that no one goes and removes it in response.

rfcs/011-dedicated-renku-git-user/011-dedicated-renku-git-user.md
jachro: Yes, only recently we were talking about that with Eike again and we ended up with the same conclusion. We have to say we don't feel comfortable adding a single user to all the projects. Imagine the credentials get compromised. I cannot even think of an emergency plan we should follow in such a scenario (I guess we'd need a very carefully prepared plan for the incident). The thing is that if the credentials leak the person would be able to do whatever: log in, use the API and access the repos, remove the original members and so on. The list is almost indefinite. The Project Access Token solution with all the known issues is in my eyes far less problematic. The access token is known only to token-repository and cannot be even viewed by the owner of the project. You cannot use it to log in, either.

Obviously, renkulab.io it's maybe a bit less of an issue as we don't guarantee anything in a written form (as far as I know). But on envs like renku health or BIT? Would it be even easy to communicate that to the responsible people?

Also, what's maybe worth mentioning is the procedure of creation of the super renku user. It might be that we'll need to cede that to the admins with some very explicit message about the importance and power of this user. I wouldn't be surprised if there will be questions about why.

rokroskar: @jachro I'm not sure I follow. The renku user wouldn't be the admin or maintainer of those projects. It would have the lowest possible read-only permissions.

wesjdj: We would like to visit this topic. Personally I think the risk of credentials leaking is the same for a single user at it is for 20k users.

PR #22, opened 2023-01-17
docs(rabbitmq): motivation for introduction of rabbitmq

seanrmurphy:
There has been much discussion of message queues within Renku and it was decided that RabbitMQ should be used as the messaging solution. However, little was recorded regarding the specific motivation for introducing this component and the choice of which messaging solution to use.

This PR captures some of that discussion, albeit significantly disconnected from many of the discussions which took place on this topic, both in time and space.


Panaetius:
probably need to change the folder to something other than 010, gateway refactor is already 010

rokroskar:
@Panaetius I believe we can close this - with existing implementation of Redis streams it isn't necessary anymore?

PR #25, opened 2023-03-20
resource control adegano

aledegano:
- feat: resource access control rfc
- grammar pass
- add notes
- [wip] refactoring
- fill last 3 sections and remove input YAML as it is confusing and will not be used

### DIFF
--- a/rfcs/011-resource-access-control/011-resource-access-control.md
+++ b/rfcs/011-resource-access-control/011-resource-access-control.md
@@ -62,230 +62,7 @@ to associate user sessions to resource pools.
 The service exposes an HTTP API that the administrators interact with to create/update/delete resource pools and
 add/remove users to resource pools.
 
-A prototype of the API exposed by the service looks like the following:
-```yaml
-openapi: 3.0.2
-info:
-  title: Renku user service
-  description: User service with compute resource access control for Renku
-  version: v1
-servers:
-  - url: /api/compute
-paths:
-  
-  /resourcePools:
-    get:
-      summary: Get all resource pool definitions
-      responses:
-        '200':
-          description: The resource pool definition
-        '404':
-          description: The resource pool does not exist
-      tags:
-        - resourcePools
-    post:
-      summary: Create a resource pool
-      requestBody:
-        required: true
-        content:
-          application/json:
-            schema:
-              type: object 
-              required:
-                - id
-              properties:
-                id:
-                  type: string
-      responses:
-        '201':
-          description: The resource pool was created
-      tags:
-        - resourcePools
-  /resourcePools/{resourcePoolId}:
-    get:
-      summary: Get a resource pool definition
-      parameters:
-        - in: path
-          name: resourcePoolId
-          required: true
-          schema:
-            type: string
-      responses:
-        '200':
-          description: The resource pool definition
-        '404':
-          description: The resource pool does not exist
-      tags:
-        - resourcePools
-  /resourcePools/{resourcePoolId}/classes:
-    get:
-      summary: Get all resource classes in a resource pool
-      parameters:
-        - in: path
-          name: resourcePoolId
-          required: true
-          schema:
-            type: string
-      responses:
-        '200':
-          description: The resource class in the specific pool
-        '404':
-          description: The resource pool or class does not exist
-      tags:
-        - classes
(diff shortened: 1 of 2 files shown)


rfcs/011-resource-access-control/011-resource-access-control.md
```diff
@@ -6,136 +6,328 @@
...
-Store information about resource pools in postgres. Create a dedicated service which will
-expose an API that will allow admins to create resource pools, modify them or assign users
-to different resource pools.
+### The Resource Access Control Service
+
+The service exposes an HTTP API that the administrators interact with to create/update/delete resource pools and
+add/remove users to resource pools.
+
+A prototype of the API exposed by the service looks like the following:
+```yaml
```
olevski, line 66: Can we please add this as a separate file in the same folder as the markdown file? Then we can simply refer to it via a link or call it out. There is no reason to embed the whole spec directly in the already fairly long markdown file.

rfcs/011-resource-access-control/011-resource-access-control.md
```diff
@@ -6,136 +6,328 @@
...
 
-In almost all cases it can happen that a session is scheduled on an untainted node/resource pool that is not strictly
-related to the resource pool that the user was "targeting". In this case the request quota from the targeted resource
-pool is consumed.
-
-### Resource pool definition
-
-This is a YAML representation of the data that would be stored in postgres.
-
-```yaml
```
olevski, line 91: I would like to keep some representation of the data that will be stored in postgres. Why do you want to fully remove this?

olevski (changes requested):
@aledegano thanks for finishing the RFC and cleaning things up. A few minor changes and comments. 

Let me know what you think.

olevski:
If you do not like the yaml representation of the postgres data then here is a bullet point list that we can use instead.

quota:
- cpu
- memory
- gpu
- storage

resource pool:
- name
- users (list of keycloak user IDs that have access to the specific resource pool)
- classes (list of references to a resource class or a list of resource classes)
- quota (reference to a quota or the content of quota)

resource class:
- name
- cpu
- memory
- gpu
- storage

Panaetius:
> resource pool:
> 
>     * name
> 
>     * users (list of keycloak user IDs that have access to the specific resource pool)
> 
>     * classes (list of references to a resource class or a list of resource classes)
> 
>     * quota (reference to a quota or the content of quota)

I think users <> resource-pool is a many-to-many relation and as such needs a separate mapping table

rfcs/011-resource-access-control/011-resource-access-control.md
```diff
@@ -149,43 +341,41 @@ flowchart TD
...
 
-> What parts of the design do you expect to resolve through the implementation 
-of this feature before stabilisation?
+The programming language chosen for the HTTP service and its dependencies will be chosen at the moment of implementation.
+At the time of implementation, some considerations on the required network policies so that the service can access Postgres
+will be needed.
 
-> What related issues do you consider out of scope for this RFC that could be 
-addressed in the future independently of the solution that comes out of this RFC?
+The problem addressed by this RFC is somewhat related to the topic of cross-cluster resource sharing.
```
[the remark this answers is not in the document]
aledegano: Generally speaking it is possible for any Kubernetes cluster: when you setup `kubelet` it only wants to know where to reach the controlplane, it does not make a distinction between local network or otherwise.

It has of course several caveats, for once the kubelet and the controlplane must be reachable in both ways, then there might be some latency issue and finally also some security concerns. Using a VPN could address many of the previous concerns.

This is also what I think about when speaking of cross-cluster resource sharing, at least in its simplest implementation.

In any case this is meant to be just a vague direction, I don't claim that I have the full design proposal for that at this moment :grin:

rfcs/011-resource-access-control/011-resource-access-control.md
aledegano: I have added a section with a possible structure for the DB.

rfcs/011-resource-access-control/011-resource-access-control.md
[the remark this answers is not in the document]
olevski: Ok cool just wanted to make sure we are on the same page. Sounds good.

olevski (approved):

PR #26, opened 2023-03-22
Add session ssh support

seanrmurphy:


aledegano (commented):
I think this RFC is very good and makes a good case for the MITM proxy.

I would spend a few words in the last section "Unresolved questions" about the "Renku data store" and how we could solve this together with the Resource Access Control RFC and its requirement for a backend.

rfcs/012-add-session-ssh-support/012-add-session-ssh-support.md
```diff
@@ -0,0 +1,438 @@
...
+developed which is currently in operation. That solution, however, has some
+limitations; this RFC provides context, documents possible solutions and will
+be used to justify the final agreed solution.
+
+## Motivation
+
+> Why are we doing this? What use cases does it support? What is the expected
+outcome?
+
+There is a [Shape-up
```
rokroskar, line 24: we've now transferred this into this repo so it can be referenced and public: https://github.com/SwissDataScienceCenter/renku-design-docs/blob/main/feature-pitches/003-ssh-into-sessions/ssh-into-sessions.md

rfcs/012-add-session-ssh-support/012-add-session-ssh-support.md
```diff
@@ -0,0 +1,438 @@
...
+The basic problem is to provide low friction, secure access to user sessions
+via ssh. 
+
+Although non key based mechansisms are possible and are discussed to some extent
+below, VS Code primarily uses key based mechansims; they are generally the most
+widely used with ssh and are known to be trusted and secure - for this reason,
+there is a strong focus on key-based mechanisms here.
+
+From a user perspective, the following needs to be considered:
+- key setup
```
rokroskar, line 42: can we simply insist on being in control of the keys? If the setup is automatic I don't see much of a need for supporting all the possible options. afaik other solutions of this type (e.g. gitpod, codespaces) don't give you any options regarding this and I never felt like this was limiting somehow. I think we _have_ to do the automatic setup anyway and then it might just save us a lot of time (at least in the first stab at this) to call it good at that point.

rfcs/012-add-session-ssh-support/012-add-session-ssh-support.md
```diff
@@ -0,0 +1,438 @@
...
+protocol operates on an IP address basis (i.e., it has no specific server
+identification as noted above). This solution can suffer from DNS propagation
+delays, depending on the DNS provider and DNS configuration. Further, exposing
+ssh servers running inside user sessions directly to the Internet brings some
+security risks and it would need to be clear that users cannot easily change
+the ssh daemon configuration within their sessions to make their session very
+exposed.
+
+A couple of further points relating to this approach:
+- ingresses on cloud providers can typically be expensive; having one per
```
rokroskar, line 170: do you mean here specifically a load-balancer ingress? When I read "ingress" I think the k8s ingress, which may or may not be a load-balancer.

rfcs/012-add-session-ssh-support/012-add-session-ssh-support.md
```diff
@@ -0,0 +1,438 @@
...
+
+The most natural solution in this case is to use the user's public key within
+the session and have open access with no ability to obtain a shell within the
+jumphost. Key management then becomes a user responsibility, with users having
+to do this within their Renku projects; as such, keys are project entities,
+rather than user entities.
+
+In this case, the command to ssh into a session requires specifying both the
+jumphost and the destination; hence it is a more complex command and introduces
+some user friction when accessing sessions. VSCode supports such proxying
```
rokroskar, line 200: Might be worth noting here that the current solution tries to solve this by encoding all the complexity in the ssh config that is generated for the user. So the user doesn't need to be aware of it at all, they issue a single ssh command to a hostname.

rfcs/012-add-session-ssh-support/012-add-session-ssh-support.md
```diff
@@ -0,0 +1,438 @@
...
+cluster ingress configuration to forward TCP traffic incident on a specific
+port to the jumphost pod on its ssh port. For cloud providers, a new ingress
+pointing at the ssh proxy is straightforward. ssh proxying approaches require
+authentication against both the proxy and the destination server; this
+necessitates an approach in which some combination of no credentials and a
+keypair must be deployed to the proxy and/or user session.
+
+The most natural solution in this case is to use the user's public key within
+the session and have open access with no ability to obtain a shell within the
+jumphost. Key management then becomes a user responsibility, with users having
```
rokroskar, line 194: Not necessarily - we could inject ssh keys into the session for the user, once there is a possibility for us to store the public keys.

rfcs/012-add-session-ssh-support/012-add-session-ssh-support.md
```diff
@@ -0,0 +1,438 @@
...
+
+Unlike the above approach, this requires writing and maintaining a new software
+component. The key difference is that this proxy does not use the standard ssh
+proxying capabilities provided by an ssh client (in which one ssh session is
+essentially embedded inside another); effectively this approach concatenates
+two distinct ssh sessions with two distinct authentication flows. The proxy
+terminates one ssh session, relaying all traffic to another ssh session.
+
+This solution has the following benefits:
+- keys can be clearly bound to users and separated from projects
```
rokroskar, line 217: Again, see my point above - the current implementation with the jumphost does not technically require that the keys are inside the project; it was simply the only way to do it with "what we have".

rfcs/012-add-session-ssh-support/012-add-session-ssh-support.md
```diff
@@ -0,0 +1,438 @@
...
+- it is reasonably straightforward to add instrumentation such that ssh session
+  information data can be easily collected with the current data collection
+  mechanisms.
+- there is greater control over the ssh command which users can use - the mapping
```
rokroskar, line 221: Maybe one additional point here - in this scenario, we can minimize the user friction further by being able to respond to ssh requests, e.g. launch a session if one doesn't yet exist.

rfcs/012-add-session-ssh-support/012-add-session-ssh-support.md
```diff
@@ -0,0 +1,438 @@
...
+|------------------------------------------------|--------------------------------------|-------------------|--------------------|-------------------|
+| VS Code support                                | :heavy_check_mark:                   | :heavy_check_mark: | :heavy_check_mark: | :x: |
+| Efficient IP address usage                     | :x:                                  | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
+| Keys linked to user rather than project        | :x:                                  | :x:                | :heavy_check_mark:  | N/A |
+| Flexible monitoring support                    | :x:                                  | :x:                | :heavy_check_mark:  | :heavy_check_mark:                |
+| Key management responsibilities with users     | :heavy_check_mark:                   | :heavy_check_mark: | :heavy_check_mark:  | :x:                |
+| New Renku components to be developed/supported | :x:                                  | :x:                | :heavy_check_mark:  | :heavy_check_mark: |
```
rokroskar, line 299: I would consider needing to develop new components a negative trait, so use the red X? :)

rfcs/012-add-session-ssh-support/012-add-session-ssh-support.md
```diff
@@ -0,0 +1,438 @@
...
+and the ssh challenge is successful using one of the user's registered ssh
+public keys.
+
+### Session creation
+
+The standard approach to session creation occurs in which the Renku notebooks
+service creates a `JupyterServer` CRD.
+
+In this case, however, a Kubernetes watcher listens for creation and deletion
+of these CRDs and updates a database accordingly. This database keeps track of
```
rokroskar, line 333: I had to read this a few times to understand that this is actually part of the proposed change, i.e. that currently no such database exists. Maybe make that a bit more explicit.

rokroskar (commented):
Thanks for putting this together @seanrmurphy! It's very useful to have all the options laid out like this. After reading through it, I can see that the MIMP is ultimately the lowest-friction solution, but it comes at a cost of implementing and maintaining a new service. There are, however, other components in this design that would already make the jumphost solution much better, namely the Renku Data Store and associated APIs + supporting services. My take on this is that we could start by implementing those pieces first since they can also be used more widely in the platform, and then re-assess the need to swap out the jumphost for the MIMP.

rfcs/012-add-session-ssh-support/012-add-session-ssh-support.md
seanrmurphy: I was just trying to itemize the different possibilities I see; definitely agree that it makes sense not to bite off all at once.

Regarding us being in control of the keys - I guess you simply mean that renku generates a keypair and the private key stays with the user.

My pref would be for a default behaviour which simply checks if a renku key exists and generates one if not, but if a flag/switch is provided, an a priori generated key can be used.

rfcs/012-add-session-ssh-support/012-add-session-ssh-support.md
seanrmurphy: Good point - I will clarify.

rfcs/012-add-session-ssh-support/012-add-session-ssh-support.md
seanrmurphy: Sure - will amend accordingly.

rfcs/012-add-session-ssh-support/012-add-session-ssh-support.md
seanrmurphy: Good point - thanks.

rfcs/012-add-session-ssh-support/012-add-session-ssh-support.md
rokroskar: sure, that makes sense. On the spectrum from "required" to "not at all needed" I would then make the automatic key generation "required" and the others "nice-to-have" (i.e. if easy or with enough time left, do it on the first implementation, otherwise defer to subsequent improvements.

rokroskar:
Thanks @seanrmurphy for the update on this - I don't have any other substantive comments, other than somehow we need to decide if implementing a new service is worth the upkeep cost. 

My personal take is that we should implement the key handling part no matter what; that is somehow a common piece between the current (interim) solution and the MITM approach suggested here. The majority of the user-facing part has already been implemented in the CLI; we would need to swap out the commit+push of a public key with an API call to the service. There are also some checks in place to determine if a project is "ssh-ready" which need to be adapted, but that seems pretty trivial. 

Once that is done, the primary difference (in my view, please correct if that's wrong) between the MITM solution and the current one is the monitoring. What kind of monitoring would we want to have in this case? Can some of it be achieved by tweaking logging on the jump server? If not, how much do we care? If the answer is that we care a lot, then the MITM approach is probably worth the effort. 

Maybe one other aspect to consider (I'm not sure exactly where it fits) is that we want to transition to having a single session per user per project (right now a user can launch many sessions from different commits of any project). This simplifies the session naming _a lot_ and would mean that the ssh configs are basically created once and left alone (in the case of the current jumphost implementation) and in the case of MITM proxy, the routing is also simplified.

rokroskar:
I found another reason for using the MITM approach - using `rclone` for filesystem access (it doesn't use standard ssh so it doesn't respect the ssh config and does not have jump proxy support)

PR #27, opened 2023-03-27
feat: add ssh-into-sessions pitch

rokroskar:
This has already been built and included in [renku 0.24.2](https://github.com/SwissDataScienceCenter/renku/releases/tag/0.24.2).

I edited the original text a bit to make it less SDSC-specific.


lokijuhy (approved):
Cool beans! FYI I pushed a few edits, mostly cleaning up some Notion formatting issues. I also deleted the screenshots and notes at the end, which were notes from the Build Team, not the pitch.

PR #28, opened 2023-05-05
feat: RFC for improving session persistence

mana-alisafaee:
Related pitch: https://www.notion.so/Make-autosaves-not-be-so-bad-120264b524ce45d485d407c14a597fb6


rokroskar:
This is great, thanks for synthesizing the shape-up @mohammad-alisafaee. I know we have "environment handling" in the rabbit-holes, but it seems to me that this might be the main reason why someone would want to launch a session from a different / older commit. The repo state can always be changed with git - the environment not so much.  Would it be possible to check at least if the files relevant for the environment configuration are different between the requested version and the running version? This is what causes users the most headaches now with the auto-save ("I installed a package, launched my session and the package wasn't there!") and it looks like this approach won't necessarily fix it?

mohammad-alisafaee:
Thanks @lorenzo-cavazzi and @rokroskar for the feedback! I've added your suggestions in the "Nice to Have" section.

lorenzo-cavazzi (dismissed):
Thank you for adding details on the different statues. Lgtm now!

rfcs/012-session-persistence/012-session-persistence.md
```diff
@@ -0,0 +1,303 @@
...
+
+### Additional Technical Details
+
+The main difference that we would introduce in Amalthea/Sessions is the following:
+
+1. Sessions exist in one of 3 forms:
+   - **Running**: This is the same as the current sessions
+       - Containers: `oauth2-proxy`, `git-proxy`, `sidecar`, `session`
+       - No changes from currently running sessions
+   - **Hibernate**:
```
leafty, line 131: Wording:
```suggestion
   - **Hibernating**:
```
and everywhere else.

leafty, line 131: Alternative wording: "Sleeping".

rfcs/012-session-persistence/012-session-persistence.md
```diff
@@ -0,0 +1,303 @@
...
+
+2. Shutting down a session (i.e. the user clicking the button in the UI to stop a session) results always in simply
+   going to the **Hibernate** form. In the case that the PV state is fully synced in git, we could consider removing the PV;
+   keeping it around means, however, that we can start the session quicker next time (e.g. the data is already fetched).
+
+3. Culling active sessions turn them into **Hibernate** form. We can consider reducing the culling interval since part of
+   the motivation for keeping it long is to avoid too many autosave branches.
+ 
+4. Additional logic will cull the **Hibernate** forms with a separate (longer) culling interval.
+
```
leafty, line 156: I would add a diagram representing the session lifecycle for clarity:

```mermaid
graph TD
  no-session(No Session)
  starting(Starting)
  running(Running)
  stopping(Stopping)
  hibernating(Hibernating)
  removing(Removing)
  deleted(Deleted)
  error(Error)

  no-session --> |start| starting
  starting --> running
  starting --> |error| error
  running --> |stop| stopping
  running --> |culled| stopping
  stopping --> hibernating
  stopping --> |error| error
  hibernating --> |start| starting
  hibernating --> |remove| removing
  hibernating --> |culled| removing
  removing --> |error| error
  removing --> deleted
```

rfcs/012-session-persistence/012-session-persistence.md
mohammad-alisafaee: I think hibernating fits very well in this context and most people liked it. `Sleeping` is also fine but let's stick to the word that we agreed upon.

rfcs/012-session-persistence/012-session-persistence.md
mohammad-alisafaee: Thanks for the diagram. I've modified it a bit and added it to the RFC.

mohammad-alisafaee:
Thanks @leafty for your comments! I've addressed them. Please take a look.

rfcs/012-session-persistence/012-session-persistence.md
```diff
@@ -0,0 +1,332 @@
...
+- Message 2: You launched a session earlier on this project, and it contains new unsaved work
+
+  - `Connect to the session` and continue from there
+  - `Discard the changes` and pick the last commit on `<branch>`
+
+- Message 3: You launched a session earlier on this project and the state has diverged from the remote repository
+
+  - `Connect to the session` and continue from there
+  - `Discard the changes` and pick the last commit on `<branch>`
```
leafty, line 257: I would strongly suggest to drop the logic around `git` state entirely from this RFC.

As mentioned above:

> From the user's perspective, connecting to a session, quitting, and reconnecting again is like coming and going from a **workspace** - they don't think about it in terms of the underlying git repository state, and we should not expect them to.

Users should be offered two choices:
1. Connect to the running or hibernating session (if possible indicate when was the last time they worked on it).
2. Discard the session and lose any unsaved work.

This creates a simple mental model around sessions for users and also reduces the implementation effort compared to the current proposal.

mohammad-alisafaee: I think it's ok to allow some notion of git in the RFC. Renku users are familiar with git anyhow and having some simple operations (e.g. fast-forwarding) should be easily understandable by them. In some places in the RFC, we talk about showing a diff of changes to users (as a nice-to-have), so, we have references to git anyway. Let's discuss this in a review meeting later.

olevski (approved):

rfcs/012-session-persistence/012-session-persistence.md
rokroskar: I think this has been addressed in the implementation.

PR #30, opened 2023-05-19
feat: RFC for GitLab spam checker

wesjdj:
A description of what a GitLab spam management service would look like.

### DIFF
--- a/rfcs/013-gitlab-spam-checker/013-gitlab-spam-checker.md
+++ b/rfcs/013-gitlab-spam-checker/013-gitlab-spam-checker.md
@@ -1,0 +1,176 @@
+# Implementing a Spam Management Service for GitLab
+
+- Start Date: 2023-05-19
+- Status: Proposed
+
+## Summary
+
+This proposal aims to introduce a spam management service to mitigate spam on GitLab and improve user experience and security. This service will use machine learning models to detect and eliminate spam across the platform in real-time where possible, focusing on snippets, issues, notes, projects, and user profiles.
+
+## Problem Statement
+
+GitLab, a service on which RenkuLab relies, is not immune to the issue of spam. Spam can infiltrate various platform aspects, including snippets, issues, notes, projects, and user profiles. Spam disrupts the user experience, poses potential security risks, and damages the reputation and RenkuLab.io's SEO. GitLab lacks a robust, real-time solution to identify and eliminate spam across all attack vectors.
+
+## Key Assumptions
+
+- The proposed machine learning model will be built in consultation with the innovation team. The machine learning model's specifics are out of this RFC's scope.
+
+## Possible Solutions
+
+- A service that only checks for spam at regular intervals. This solution is less than ideal because of the delay between the spam being posted and the spam being detected and removed, during which spam could already be harmful. 
+
+- Monitor GitLab for spam manually. This solution is unsuitable because it is time-consuming.
+
+- GitLab has its [Spamcheck](https://docs.gitlab.com/ee/user/admin_area/reporting/spamcheck.html) service, which can be used to check for spam issues and snippets. This service does not check for spam in notes, projects, or user profiles. Currently, it is not possible to configure the service to delete spam automatically. Deploying GitLab's Spamcheck service also requires GitLab Enterprise Edition, which RenkuLab.io's GitLab is not currently running on.
+
+- GitLab can make use of [Akismet](https://akismet.com/), a third-party-hosted AI spam filter. However, this only checks for spam on issues and notes, needs training, and would involve handing data over to a third party in the US.
+
+## Proposed Solution
+
+The proposed spam management service will integrate with GitLab via system hooks and regularly scheduled jobs. When notes, projects, or users are created or updated, GitLab will send system hooks to a spam management service. System hooks request examples can be found [here](https://docs.gitlab.com/ee/administration/system_hooks.html#hooks-request-example).
+
+System hooks triggered on group, project, and user events can be triggered at the system level. Issues and comments at a project level would require a project-level system hook. This could be configured when a project is added to KG, or KG could forward project-level events to the spam management service's event service. Project-level webhook events examples can be found [here](https://docs.gitlab.com/ee/user/project/integrations/webhook_events.html).
+
+Snippets do not have system hooks. Therefore a job triggered at regular intervals will be required to check for new snippets. This can be done by checking the snippet API for new snippets. Documentation on the snippet API can be found [here](https://docs.gitlab.com/ee/api/snippets.html).
+
+Ideally, the solution will also be able to check abuse reports manually submitted by users in GitLab and check the reported URL in the abuse report for spam content.
+
+A feedback mechanism would also be implemented, allowing the machine learning model to learn from mistakes and improve over time. This could be done by storing content in an S3 bucket and regularly re-training the model on a dataset including new data. The system should also keep a log of decisions made and the confidence level of the decision, as well as export Prometheus metrics which can be used to compile regular reports.
+
+Similarly to GitLab's Spamcheck service, storing the model in an S3 bucket and pulling it to the spam management service on start-up would allow the model to be updated without having to redeploy the spam management service.
+
+An allowlist of trusted email domains could be implemented to prevent false positives and ensure users with trusted email domains can post content without being checked for spam. YAT already possesses a list of trusted email domains. Users without trusted email domains could also be verified with a 'verified' string in the GitLab user's admin note. A list of verified users would be stored in Redis and updated periodically. This would allow users to be verified without having to redeploy the spam management service.
+
+A queue management system such as [Redis lists](https://redis.io/docs/data-types/lists/) would be used to track items that need to be classified and spam items that need to be deleted from GitLab. This helps to ensure the system can handle many system hooks from GitLab without significant latency and without overloading the GitLab API. Redis was chosen because of GitLab's existing reliance on Redis; and not wanting to add another dependency to the system, which should be able to work independently of RenkuLab. Depending on the amount of data sent to the service and its efficiency, it may be necessary to implement Redis persistence in the case of a crash.
+
+This mermaid diagram provides an overview of how this system would work:
+
+```mermaid
+  flowchart TB
+    subgraph "GitLab Components"
+      GH[GitLab System Hook]
+      API[GitLab API]
+    end
+
+    subgraph "Redis Components"
+      Q1[GitLab Item Fetch Queue]
+      Q2[User Trust and Verification Queue]
+      CQ[Classification Queue]
+      DQ[Deletion Queue]
+      VRU[Verified Renku Users List]
+    end
+
+    subgraph "Spam Management System"
+      subgraph "Jobs"
+        SC[Snippet Check Job]
+        VU[Verified Users Retriever Job]
+      end
+      subgraph "Services"
+        ES[Event Service]
+        FS[GitLab Item Fetch Service]
+        TV[User Trust and Verification Service]
+        CS[Classification Service]
+        DS[Deletion Service]
+        subgraph "Other Components"
+          S3[S3 Bucket]
+          L[Logs]
+          TED[Trusted Email Domains List]
+          ML[Classifier Model]
+        end
+      end
+    end
+
+    GH -->|Triggered on item create/update| ES
+    SC -->|Runs Periodically| ES
+    VU -->|Updated Periodically| VRU
+    ES --> Q1
+    Q1 --> FS
+    FS -->|Get Item| API
+    FS -->|Add Item To Queue| Q2
+    Q2 --> TV
+    TED --> TV 
+    VRU --> TV 
+    VU -->|Get Verified Users| API
+    TV -->|If Not Trusted Or Verified| CQ
+    CQ --> CS
+    CS -->|If Item Is Spam| DQ
+    CS -->|Classification Result| L
+    DQ --> DS
+    DS -->|Delete Item from GitLab| API
+    DS -->|Set Item to Private in GitLab| API
+    CS -->|Store Evaluated Content| S3
+    S3 -->|Download Model On Startup| ML
+    CS <-->|Classification| ML
+    
+    style GH fill:#E24329,stroke:#333,stroke-width:4px
+    style API fill:#E24329,stroke:#333,stroke-width:4px
+    style ES fill:#645EB6,stroke:#333,stroke-width:4px
+    style SC fill:#E46991,stroke:#333,stroke-width:4px
+    style FS fill:#645EB6,stroke:#333,stroke-width:4px
+    style TV fill:#645EB6,stroke:#333,stroke-width:4px
+    style CS fill:#645EB6,stroke:#333,stroke-width:4px
+    style DS fill:#645EB6,stroke:#333,stroke-width:4px
+    style VU fill:#E46991,stroke:#333,stroke-width:4px
+    style Q1 fill:#009A6A,stroke:#333,stroke-width:4px
+    style Q2 fill:#009A6A,stroke:#333,stroke-width:4px
+    style CQ fill:#009A6A,stroke:#333,stroke-width:4px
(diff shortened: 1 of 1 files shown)


rfcs/013-gitlab-spam-checker/013-gitlab-spam-checker.md
```diff
@@ -0,0 +1,93 @@
+# Implementing a Spam Management Service for GitLab
+
+- Start Date: 2023-05-19
+- Status: Proposed
+
+## Summary
+
+This proposal aims to introduce a spam management service to mitigate spam on GitLab and improve user experience and security. This service will use machine learning models to detect and eliminate spam across the platform in real-time, focusing on snippets, issues, notes, projects, and user profiles.
```
seanrmurphy, line 8: I think it would make sense to quality real-time here; further down, you mention a mix of hook based mechanisms plus mechanisms which are schduled, so I guess it's not all real-time; also I guess some wil require human intervention in which case the detection cuold be somehow real-time but the elimination might not be.

rfcs/013-gitlab-spam-checker/013-gitlab-spam-checker.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Problem Statement
+
+GitLab, a service on which RenkuLab relies, is not immune to the issue of spam. Spam can infiltrate various platform aspects, including snippets, issues, notes, projects, and user profiles. Spam not only disrupts the user experience but also poses potential security risks. GitLab lacks a robust, real-time solution to identify and eliminate spam across all attack vectors.
```
seanrmurphy, line 12: Spam is also detrimental to reputation.

rfcs/013-gitlab-spam-checker/013-gitlab-spam-checker.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Key Assumptions
```
seanrmurphy, line 14: I guess a key assumption is admin access to a gitlab api is available?

rfcs/013-gitlab-spam-checker/013-gitlab-spam-checker.md
```diff
@@ -0,0 +1,93 @@
...
+
+- The proposed machine learning model will be built in consultation with the innovation team. The machine learning model's specifics are out of this RFC's scope.
+
+## Possible Solutions
```
seanrmurphy, line 18: I think it would be good here to itemize the options in bullets.

 I also think it would be good to highlight that the problem is quite specific and eg email spam detection mechanisms can not be readily applied as the natuire of the content is quite different...

rfcs/013-gitlab-spam-checker/013-gitlab-spam-checker.md
```diff
@@ -0,0 +1,93 @@
...
+System hooks triggered on group, project, and user events can be triggered at the system level. Issues and comments at a project level would require a project-level system hook. This might be able to be configured when a project is added to KG. Project-level webhook events examples can be found [here](https://docs.gitlab.com/ee/user/project/integrations/webhook_events.html).
+
+Snippets are unable to trigger system hooks. Therefore a job triggered at regular intervals will be required to check for new snippets. This can be done by checking the snippet API for new snippets. Documentation on the snippet API can be found [here](https://docs.gitlab.com/ee/api/snippets.html).
+
+Ideally, the solution will also be able to check abuse reports manually submitted by users in GitLab and check the reported URL in the abuse report for spam content.
+
+A feedback mechanism would also be implemented, allowing the machine learning model to learn from mistakes and improve over time. This could be done by storing content in an S3 bucket and regularly re-training the model on a dataset including new data. The system should also keep a log of decisions made and the confidence level of the decision.
```
seanrmurphy, line 36: This feels somehow like the second phase of the work; the objective in the near term would be to get a variant operational which has no specific feedback mechanisms?

rfcs/013-gitlab-spam-checker/013-gitlab-spam-checker.md
```diff
@@ -0,0 +1,93 @@
...
+
+Similarly to GitLab's Spamcheck service, storing the model in an S3 bucket and pulling it to the spam management service on start-up would allow the model to be updated without having to redeploy the spam management service.
+
+An allowlist of trusted email domain providers could be implemented to prevent false positives to ensure users with trusted email domains can post content without it being checked for spam. YAT already possesses a list of trusted email domains.
+
+This mermaid diagram provides an overview of how this system would work:
```
seanrmurphy, line 42: I find the diagram a bit confusing....it looks like the spam checking service 'checks' different endpoints, but it's not clear what this check consists of...does it mean that it checks for new content/changes since the last check took place? Does it retrieve the content and aggregate it and send it to the classifier? Would it make sense to make the box containing the spam checking service larger such that it contains at least the model and perhaps a cleanup function and maybe some scheduler for checking issues which need to be proactively checked?

rfcs/013-gitlab-spam-checker/013-gitlab-spam-checker.md
```diff
@@ -0,0 +1,93 @@
...
+    linkStyle 8 stroke:#2ecd71,stroke-width:2px;
+    linkStyle 9 stroke:#2ecd71,stroke-width:2px,dasharray: 5, 5;
+    linkStyle 10 stroke:#2ecd71,stroke-width:2px;
+    linkStyle 11 stroke:#2ecd71,stroke-width:2px;
+    linkStyle 12 stroke:#2ecd71,stroke-width:2px;
+```
+
+## Drawbacks
+
+1. Ensuring the system can handle many system hooks from GitLab without significant latency is critical. It needs to be scalable and able to handle peak loads.
```
seanrmurphy, line 78: Meaning it should not impact the performance of gitlab, right? I think this should be made more explicit...

rfcs/013-gitlab-spam-checker/013-gitlab-spam-checker.md
```diff
@@ -0,0 +1,93 @@
...
+2. The spam management service will require a GitLab personal access token with administrative rights to delete spam.
+3. We must ensure we can be informed of and restore incorrectly classified content.
+
+## Rationale and Alternatives
+
+This solution is deemed the best as it provides real-time spam detection without intruding on the user experience. Alternatives such as manual monitoring are too time-consuming.
+
+Not implementing this solution would mean continuing to expose GitLab users to potential spam, which could disrupt their experience and pose security risks, and open RenkuLab.io up to promoting unethical content.
+
+## Unresolved Questions
```
seanrmurphy, line 88: How to calibrate it such that it can be trusted to act sensibly on obvious spam, and to flag possible spam to administrators without having too many false positives...

rfcs/013-gitlab-spam-checker/013-gitlab-spam-checker.md
wesjdj: For example, 80 issues can sometimes be added in a matter of seconds. We need to make sure the spam management service is available to accept that number of hooks in that time frame, otherwise, some might fall between the gaps

rfcs/013-gitlab-spam-checker/013-gitlab-spam-checker.md
wesjdj: The spam management service shouldn't require human intervention. Where possible, hooks are used so its real time, and where GitLab doesn't supply hooks, such as for snippets, we'll use jobs at regular interviews

rfcs/013-gitlab-spam-checker/013-gitlab-spam-checker.md
wesjdj: Well, we don't know about that. We might try building on email spam detection models. Its out of the scope of this RFC though.

rfcs/013-gitlab-spam-checker/013-gitlab-spam-checker.md
wesjdj: I don't think so - we need to be able to log what decisions are made and why from the start. Our current hacked together spam management scripts can do this so I see no reason why the service shouldn't be able to as well.

rfcs/013-gitlab-spam-checker/013-gitlab-spam-checker.md
wesjdj: I don't want to make too many assumptions about the design of the actual service yet so I'm trying to keep it quite high level. Think of the upper items as inputs/triggers to the service and the lower items as actions/outputs

rfcs/013-gitlab-spam-checker/013-gitlab-spam-checker.md
wesjdj: It could be better but its abstract for a reason atm, I will evolve it as the design of the system becomes more clear

rfcs/013-gitlab-spam-checker/013-gitlab-spam-checker.md
wesjdj: Yeah, we'll take that up with innovation team because I think that will largely depend on the specifics of the model

rfcs/013-gitlab-spam-checker/013-gitlab-spam-checker.md
aledegano: oh, this is a good point to expand on the design: since we rely on events "push" (e.g. webhook incoming from GL) we need to make sure we store those in a stack and then process them pulling one by one. 
AFAIK an event based queue works well for this purpose, and the queue itself should be persisted (so if the service restarts it's not lost).

What Sean mentions, it's a different topic, but one that needs to be made explicit indeed: when the service acts on GL we should make sure that we won't hammer GL too much.

rfcs/013-gitlab-spam-checker/013-gitlab-spam-checker.md
aledegano: I'd also lean into having the feedback mechanism from the start and I would spend some words on its design.

rfcs/013-gitlab-spam-checker/013-gitlab-spam-checker.md
wesjdj: Hopefully less of an issue with GLCN but yes indeed

rfcs/013-gitlab-spam-checker/013-gitlab-spam-checker.md
```diff
@@ -0,0 +1,177 @@
...
+
+## Possible Solutions
+
+- A service that only checks for spam at regular intervals. This solution is less than ideal because of the delay between the spam being posted and the spam being detected and removed, during which spam could already be harmful. 
+
+- Monitor GitLab for spam manually. This solution is unsuitable because it is time-consuming.
+
+- GitLab has its [Spamcheck](https://docs.gitlab.com/ee/user/admin_area/reporting/spamcheck.html) service, which can be used to check for spam issues and snippets. This service does not check for spam in notes, projects, or user profiles. Currently, it is not possible to configure the service to delete spam automatically. Deploying GitLab's Spamcheck service also requires GitLab Enterprise Edition, which RenkuLab.io's GitLab is not currently running on.
+
+- GitLab can make use of [Akismet](https://akismet.com/), a third-party-hosted AI spam filter. However, this only checks for spam on issues and notes, needs training, and would involve handing data over to a third party in the US.
```
olevski, line 26: Why is Akismet suitable only for issues and notes? I thought it can check any body of text? Also I thought that as a paid service it does not need to be trained.

rfcs/013-gitlab-spam-checker/013-gitlab-spam-checker.md
```diff
@@ -0,0 +1,177 @@
...
+
+- The proposed machine learning model will be built in consultation with the innovation team. The machine learning model's specifics are out of this RFC's scope.
+
+## Possible Solutions
+
+- A service that only checks for spam at regular intervals. This solution is less than ideal because of the delay between the spam being posted and the spam being detected and removed, during which spam could already be harmful. 
+
+- Monitor GitLab for spam manually. This solution is unsuitable because it is time-consuming.
+
+- GitLab has its [Spamcheck](https://docs.gitlab.com/ee/user/admin_area/reporting/spamcheck.html) service, which can be used to check for spam issues and snippets. This service does not check for spam in notes, projects, or user profiles. Currently, it is not possible to configure the service to delete spam automatically. Deploying GitLab's Spamcheck service also requires GitLab Enterprise Edition, which RenkuLab.io's GitLab is not currently running on.
```
olevski, line 24: Do we have access to the Spamcheck code? And if yes, can we deploy it and call its API ourselves? This would mean we do not have to develop the model ourselves.

wesjdj: The Spamcheck code is [public](https://gitlab.com/gitlab-org/gl-security/security-engineering/security-automation/spam/spamcheck). The model, however, is not included in the repo and is instead downloaded from an S3 bucket when the service starts. I have done what you have suggested for evaluating existing content on our GitLab for spam. There are some improvements we can make, though. GitLab themselves have disabled the ability of Spamcheck to remove content from GitLab, presumably because they don't trust its results yet: https://gitlab.com/gitlab-org/gl-security/security-engineering/security-automation/spam/spamcheck/-/blob/v1.4.1/app/spammable.py#L26-33

rfcs/013-gitlab-spam-checker/013-gitlab-spam-checker.md
rokroskar: Partially. It doesn't need to be trained, it's learning continuously and merging all information about spam into one global model. You can contribute to its training by submitting pieces of known spam. It's limited to issues and snippets only in the way gitlab uses it, but it's made to be used for any user-entered text anywhere through its API.

rfcs/013-gitlab-spam-checker/013-gitlab-spam-checker.md
wesjdj: The main problem with Spamcheck is that it doesn't check anything besides snippets and issues. So it doesn't fulfil all of our requirements.

rokroskar: but that's a question of its integration into gitlab's functionality - the model is still available to be used to validate the events as you propose, no?

rfcs/013-gitlab-spam-checker/013-gitlab-spam-checker.md
rokroskar: aha yes - but afaik this just trains the global model, not one for you specifically. I think it's performance is dependent on how much of the additional metadata you supply.

rokroskar: I could be wrong about global vs specific - their documentation is not super obvious imho

rfcs/013-gitlab-spam-checker/013-gitlab-spam-checker.md
wesjdj: Not really - I want the model to be able to detect the presence or absence of attributes as a way of identifying spam.

wesjdj:
Some feedback from the design meeting:

- Instead of immediate deletion of users, offending items should be set to private until a backup has been made of GitLab. After that point, the offending items can be deleted. This allows for incorrectly deleted items to be restored.
- To reduce the amount of API calls to GitLab, trusted domain or verified user status could be established before the GitLab Item Retriever Service is called.
- Some Redis queues - for example, the classification queue - may be surplus to requirements.
- We should reach out to GitLab or the GitLab community to see whether there are better ways to retrieve information of events from GitLab.
- KG could forward events at a project level to the spam management service to reduce the need for a separate spam management service webhook in each project.
- Monitoring and reporting of actions taken by the spam management service should be created, such as a Prometheus exporter and a scheduled Grafana report, and a Slack webhook.
- The spam management service should allow additional custom actions when an item is deleted or disabled in GitLab. We would use this to notify the KG of the item's removal/disabling.

olevski (approved):

PR #35, opened 2024-01-25
Add Renku 1.0 Compute sessions pitch

lokijuhy:
and update roadmap


feature-pitches/010-compute-sessions/compute-sessions.md
```diff
@@ -0,0 +1,220 @@
...
+- **At a minimum, 1 default environment must be created as part of the build!** (a Jupyter Python environment)
+- These images are pre-built, so launching  a session from one is fast (the user doesn’t have to wait for an image build)
+- [nice to have] The default environments should be configured via the admin panel
+
+##### Priority 2: Use an external image
+
+- There are a whole bunch of restrictions for what must be in the image for it to work on RenkuLab. We will have to document these limitations for users.
+  - We *think* that any image coming from the Jupyter stack should work
+  - ... Or at least Renku base images
+- Only publicly accessible images are supported
```
rokroskar, line 115: This should either say "... will be supported as a part of this build, and authentication options added later." _or_ add a "nice to have" to deal with authentication.

feature-pitches/010-compute-sessions/compute-sessions.md
```diff
@@ -0,0 +1,220 @@
...
+##### Priority 3 [nice to have]: Configure a Session from a Code repo
+
+- This uses the already existing CI image build pipeline
+- We do not automatically update the image that is used for the session when new commits are pushed and new images are built. It is up to the user if they want to update which image is built
+- [nice-to-have]: The user is notified when a new image is available and invited to change the image their session uses if they wish.
+
+##### Priority 4: [for the future/out of scope] Search & select from preexisting RenkuLab environments
+
+#### User Flow: Start Session
+
```
rokroskar, line 126: there is no provision here for selecting the resource pool / class?

feature-pitches/010-compute-sessions/compute-sessions.md
```diff
@@ -0,0 +1,220 @@
...
+#### Starting Assumptions
+
+- In this build, we will not change the project GitLab CI-based image build process (for 'Configure a Session from a Code repo')
+
+#### Key Points
+
+- When a user creates a new project, the project has no sessions until a user adds one.
+  - [nice to have] Provide shortcuts for adding a session from the RenkuLab default environments
+- A user can create as many sessions as they want in a project.
+- A user can run 1 session per environment per project (similar to our restriction now of 1 session per user per project)
```
rokroskar, line 152: So we need to actually prevent the user from creating multiple sessions from the same environment - should this be mentioned in the user flow?

feature-pitches/010-compute-sessions/compute-sessions.md
```diff
@@ -0,0 +1,220 @@
...
+  /external-storage-2
+```
+
+### Revisiting the Motivating Problems
+
+- When I am starting a new project, I want to **start out with a basic compute environment** (for example with a pre-built data science stack), and I’ll decide what extra packages I need later.
+  - → Solved by offering RenkuLab built-in session “templates” (name TBD) as the fastest route to adding a session to a project
+- When I create a new Renku project, I want to get started right away with a pre-built environment- **I don’t want to have to wait for the environment image to be built** to get started.
+  - → Same as above
+- When I start a session, I want to **clearly understand what environment image the session will use and what is installed in it**, so I am not confused where the compute environment comes from.
```
rokroskar, line 182: imho this doesn't really address the "clearly understand what is installed" part?

rokroskar (commented):
Thanks @lokijuhy!! Just some points to clarify

feature-pitches/010-compute-sessions/compute-sessions.md
```diff
@@ -0,0 +1,220 @@
...
+An environment has the following properties:
+
+- name
+- an image/tag
+- metadata:
+  - who created it
+  - date of creation
+- [optional] description (e.g. why it is useful)
+- [optional] keywords/tags to aid in searching (for example, I might tag an environment with `#python` `#pytorch` to give a quick impression of what is in the image)
+
```
anciaux, line 52: In a very general sense, I feel that these changes are centered on the actual Docker images Renku is using. This is very helpful, however there is something that we cannot do with docker images: "mixing" the environments. 

Let's put a concrete case:

Let's say I have an app for "paraview" GUI viewer and an app for a finite element software and another and finally one for Solidipes. The user would wish to create a single session with /paraview, /solidipes, /fe and /lab automatically. In the current Renku, I would do with python plugins and jupyter proxy: this is conditioned by the currently installed python packages. But what is proposed here, do not allow that, I think.

How am I supposed to mix environments ? Can I create as session with two environments ? From your proposal, I think that someone has to make a Dockerfile (somewhere) and describe the installation for the mix.
However, it would make sense to split the process of installation (and not requiring the users to write the Dockerfile for any combination). 

I think the metadata attached to an environment should also include possible (compatible) requirements. That would allow the dependencies to be somehow selected while creating the session (and select them in a catalog). 

Another possibility could be to have two types of "environments": one for the Dockerfile level and one for the requirements. (Sadly, this is orthogonal to the request that users do not want to wait installation times when spawning a new session....).

To summarize: I think it is important to think of a way/procedure, which allows mixing environment/apps easily.

rokroskar: yikes, we left this unanswered for a little too long. You are totally right @anciaux that mixing environments should be possible, but I think that has to be done after this initial work (btw, this is already implemented). There are two possible options for what you are proposing:

1. pluggable front-ends (the user provides a docker image with their runtime, but another independent front-end is added to their session)
2. multi-container sessions

I think both of these have use-cases and I would like to see us work out how to support them. In the case of front-ends (option 1), they would need to be configured and maintained by the admins of the platform. The disadvantage is that this would be less flexible for the user, but probably lead to better UX because it would be more controlled and carefully managed. Option 2 gives more flexibility but the containers would not share the runtime so in the case of e.g. python or other packages needed across containers, it would not be straightforward to share them.

rokroskar (approved):

